PR #5, opened 2026-04-23
Tab-bar overflow dropdown with hide/unhide support

### LINKED ISSUE #2: Revamp the tab bar and allow hidden tabs
I want to add a new button at the far right with the three dots as a label. It will open a dropdown Inside it there will be:
* "Rename"
* "Close"
* "Hide" (or "Unhide") depending on if the tab is already hidden or not
* A separator section
* <all hidden tabs>

They will all have icons - appropriately picked ones for the "system menu items" like close (reused from other parts of the app, if possible/appropriate) and the hidden tabs will have their ordinary type-dependent icon.

A tab that is hidden is not shown on the tab strip. It's shown only in the dropdown instead.

The current functionality where if we hover to the right of the tab label to open a menu will be discarded (and so will the space that "hover icon" consumes).

soderbjorn:
Closes #2

## Summary
Replaces the old per-tab hover menu with a single global "⋮" overflow button at the far-right edge of the tab strip. The dropdown carries Rename / Close / Hide-or-Unhide actions for the active tab plus a separator and one entry per hidden tab so the user can re-activate hidden tabs from one always-reachable place. Tabs flagged as hidden are dropped from the strip rendering but still keep their content area, so an active+hidden tab keeps showing its panes.

## Background
From the issue: a new far-right three-dots button should open a dropdown containing Rename / Close / Hide-or-Unhide (state-dependent), a separator, and the list of all hidden tabs (each shown with its type-dependent icon). A hidden tab is not rendered on the strip, only in the dropdown. The previous per-tab hover-to-the-right-of-the-label menu — and the horizontal space its hover icon reserved on every tab — is to be discarded.

The issue is short on UX edge cases (e.g. what happens when the active tab gets hidden, what the menu shows for a hidden+active tab, what to do about hiding the last visible tab). The "Decisions & reasoning" section below records every interpretation I made.

## Approach
The change spans four layers:

1. **Data model (`clientServer/.../WindowConfig.kt`)** — add `hidden: Boolean = false` to `TabConfig`. Default false means existing persisted blobs deserialize as fully visible — no migration needed.

2. **Wire protocol (`clientServer/.../WindowCommand.kt`)** — new `WindowCommand.SetTabHidden(tabId, hidden)` mutation, wired through `Application.handleWindowCommand` to a new `WindowState.setTabHidden(tabId, hidden)` server function. The server refuses to hide the last visible tab and hops the active selection to the next visible neighbour when the active tab is hidden — same fallback shape as `closeTab`.

3. **Tab strip rendering (`web/.../WindowConnection.kt`)** — the strip render loop now skips tabs with `hidden == true`. The content section loop is unchanged, so an active+hidden tab still renders its panes (the active indicator just hides, since `positionActiveIndicator` already handles a missing `.tab-button.active` gracefully). The whole `run { ... }` block that built the per-tab `.tab-menu`, hover button, and floating menu list is gone, along with its CSS.

4. **Overflow menu (`web/.../OverflowMenu.kt`, new)** — a single dropdown wired up in `index.html` between the tab strip and the header actions cluster. `initTabOverflowMenu()` is called once at boot; `populateTabOverflowMenu()` rebuilds the body on every open so the hidden-tab list and the Hide/Unhide label always reflect the latest server push. Each row carries an inline-SVG icon per the issue spec (pencil for Rename, X for Close, eye / eye-with-slash for Unhide / Hide, and per-pane-kind icons for hidden tabs).

A small refactor extracts the shared leaf-icon SVG palette from `Sidebar.kt` into `iconSvgForLeafKind(contentKind, isLink)` so the sidebar tree and the new overflow dropdown render identical icons for the same pane type.

### Hidden flag lives on `TabConfig`, not in a separate UI-settings blob
Hidden is per-tab and changes transactionally with tab creation/closure; storing it inside the same `WindowConfig` JSON blob means it persists in lock-step with the rest of the tab state and survives server restarts via the existing debounced save flow. Putting it in the `ui.settings.v1` blob (which `SettingsRepository` keeps for sidebar collapse state, appearance, etc.) would have introduced a separate persistence path the client would have to keep coherent with the tab list, and the memory note "all UI settings stored server-side in SQLite only" explicitly endorses the existing tab-config blob as the right home.

### Hiding the active tab auto-switches to the next visible neighbour
The alternative is to allow `active && hidden` quietly. I went with auto-switching for the *Hide* action because:
  1. It mirrors how `CloseTab` already behaves (the user lands on a neighbour, never on nothing).
  2. It keeps the strip visually meaningful — there is always a highlighted tab unless the user explicitly activates a hidden one from the dropdown.

The active+hidden state is still reachable: clicking a hidden tab in the dropdown activates it without unhiding (so the user can preview a hidden tab without committing to restoring it). This is the only path into the active+hidden state, and the menu's Hide → "Unhide" toggle handles the exit.

### Refuse to hide the last visible tab (server *and* client)
Keeping at least one tab on the strip avoids the dead-end UX where the strip is empty and the only way to recover is the dropdown. The server rejects the command and the client also disables the menu item, so the user gets immediate feedback instead of a silent no-op. Same posture as `CloseTab` refusing to close the last tab.

### Single global overflow button vs. one-per-tab menu
The issue asks for "a new button at the far right with the three dots as a label" — singular, far-right. A per-tab variant would have brought back the very horizontal-space tax we just removed. The single button means Rename / Close / Hide always operate on the *active* tab. To rename or close an inactive tab, the user clicks it first to make it active. This is a small UX regression for editing inactive tab titles, but it keeps the strip uncluttered and matches the issue's wording.

### Overflow dropdown anchored *outside* the scrolling tab bar
The tab bar uses `flex: 1` plus horizontal scroll for overflow tabs. Placing the new button *inside* the bar would let it scroll out of view exactly when the user has lots of tabs (the case where they most want to hide some of them). Placing it in the header between the tab bar and `.header-actions` keeps it always reachable; with `align-self: flex-end` it sits flush with the tab buttons.

### Rename for a hidden+active tab falls back to `window.prompt`
`startTabRename` mutates the tab's strip label into an `<input>`. A hidden tab has no label to mutate. Three options: (a) auto-unhide first, (b) build a custom in-app dialog, (c) browser `window.prompt`. Option (a) silently changes user state (the tab unhides without an explicit ask). Option (b) is more code than the rare flow warrants. Option (c) is ugly but functional and obvious. Picked (c) — if the flow proves common in practice, an in-app dialog is a one-file follow-up.

### Per-hidden-tab icon = first-pane content kind
Tabs do not have an intrinsic type — only their panes do. The issue says hidden tabs "will have their ordinary type-dependent icon", so I picked the most natural single-icon mapping: the first pane's content kind, with `terminal` as the fallback for empty tabs. This matches what the sidebar does conceptually (icon-per-pane) compressed to one-icon-per-tab. A composite "stack of icons" would have been fancier but not asked for.

### Click a hidden tab → activate without unhiding
The issue says "It's shown only in the dropdown instead" but doesn't specify what clicking it does. Activating without unhiding lets the user *preview* a hidden tab from the dropdown without committing to restoring it (and the Hide/Unhide toggle then becomes "Unhide", giving them the explicit path back). The alternative — auto-unhide on activation — would never let the user view a hidden tab without immediately altering its hidden state.

### Refactor `iconSvgForLeafKind` rather than duplicate the SVGs
The four leaf icons (terminal / file browser / git / link) were inline literals in `Sidebar.kt`. I needed identical icons in `OverflowMenu.kt`. Pulling them into a shared function keeps the two surfaces from drifting; future changes to the icon set live in one place.

## Assumptions
- Existing persisted `WindowConfig` blobs in users' SQLite databases will deserialize cleanly with `hidden = false` (kotlinx-serialization treats absent fields as the default).
- The existing Hide/Unhide toggle on the *active* tab is the most natural place for the action; no per-tab right-click menu in the strip is required. (The issue does not mention one.)
- The horizontal space saved by removing the per-tab `.tab-menu` is welcome — the issue explicitly asks for it. No existing UX flow depends on the hover icon being there.

- **Per-tab right-click context menu** — would re-introduce a per-tab menu surface immediately after the issue asked us to remove one; not in the spec.
- **Auto-unhide a hidden tab on activation from the dropdown** — would prevent any "preview without unhiding" flow.
- **Move the overflow button inside the tab bar (right of `+`)** — scrolls out of view when there are many tabs, defeating the purpose.
- **Parallel `ui.settings.v1` storage for the hidden flags** — splits the source of truth; server would have to merge two stores on every render.

## Verification
- `./gradlew :clientServer:compileKotlinJvm :server:compileKotlin :web:compileKotlinJs` — BUILD SUCCESSFUL. Only pre-existing warnings (`@DelicateApi` notes in unrelated files).
- `./gradlew :clientServer:assemble :web:jsBrowserDevelopmentWebpack` — BUILD SUCCESSFUL on JS bundle and Kotlin/Native iOS framework links, so the data-model addition doesn't break iOS or the web bundler.
- `./gradlew :server:test` — BUILD SUCCESSFUL (the existing `StateDetectorTest` passes; no new tests added — `WindowState.setTabHidden` is small enough that the policy is enforced in a single function readable end-to-end, and the existing `WindowState` has no test scaffolding to extend).
- **Gap:** I did not exercise the dropdown end-to-end in a live browser session — I have no way to launch the server + open `localhost` in this environment. Flows that should be eyeballed before merge:
  1. Open the menu, click Hide on the active tab, confirm the strip drops the tab and the next visible tab takes over as active.
  2. Open the menu again, click the now-listed hidden tab, confirm it activates (its panes show up) and the orange ring disappears from the strip.
  3. Click Unhide, confirm the tab reappears in the strip with the ring back on it.
  4. Try to hide when only one visible tab remains — the menu item should appear greyed out.
  5. Rename + Close still work for the active tab (they go through the same server commands as before).

## Files of note
- `web/.../OverflowMenu.kt` — new file; the dropdown's controller plus the four leading-icon SVG literals (pencil, X, eye, eye-off).
- `web/.../WindowConnection.kt` — strip render now skips hidden tabs; the entire per-tab `.tab-menu` block is deleted.
- `web/.../Sidebar.kt` — `iconSvgForLeafKind()` extracted; the existing render call site is one line shorter.
- `server/.../WindowState.kt` — new `setTabHidden()` with the last-visible refusal and active-hop policy.
- `web/.../resources/styles.css` — new `.tab-overflow-*` rules; the dead `.tab-menu*` block is deleted.
- `web/.../resources/index.html` — new `<div id="tab-overflow-dropdown">` between the tab nav and `.header-actions`.

## Follow-ups
- Replace the `window.prompt` fallback for renaming a hidden+active tab with a small in-app dialog if the flow turns out to be common.
- Consider a quick "Unhide" affordance on each hidden-tab row in the dropdown (e.g. an inline icon button) so the user does not have to activate-then-unhide.
- The shared leaf-icon palette in `iconSvgForLeafKind` is a natural home for any future pane content kinds — adding one means updating just that function instead of two surfaces.

### DIFF
--- a/clientServer/src/commonMain/kotlin/se/soderbjorn/termtastic/WindowCommand.kt
+++ b/clientServer/src/commonMain/kotlin/se/soderbjorn/termtastic/WindowCommand.kt
@@ -286,6 +286,22 @@ sealed class WindowCommand {
     data class RenameTab(val tabId: String, val title: String) : WindowCommand()
 
     /**
+     * Toggle whether [tabId] is hidden from the tab strip. Hidden tabs are
+     * still listed in the tab-bar overflow dropdown so the user can recover
+     * them. The server refuses to hide the last visible tab (the strip would
+     * be empty); when hiding the active tab, the server switches the active
+     * selection to the next visible neighbour so the user lands on a tab
+     * with a strip button instead of nothing.
+     *
+     * @param tabId the id of the tab to hide or unhide
+     * @param hidden `true` to hide the tab from the strip, `false` to restore it
+     * @see TabConfig.hidden
+     */
+    @Serializable
+    @SerialName("setTabHidden")
+    data class SetTabHidden(val tabId: String, val hidden: Boolean) : WindowCommand()
+
+    /**
      * Reorder a tab by moving it before or after another tab via drag-and-drop.
      *
      * @param tabId the id of the tab being moved
--- a/clientServer/src/commonMain/kotlin/se/soderbjorn/termtastic/WindowConfig.kt
+++ b/clientServer/src/commonMain/kotlin/se/soderbjorn/termtastic/WindowConfig.kt
@@ -62,6 +62,16 @@ data class TabConfig(
      * "let the client default to the first pane in DOM order".
      */
     val focusedPaneId: String? = null,
+    /**
+     * Whether this tab is hidden from the tab strip. Hidden tabs still exist
+     * (their panes and PTY sessions are preserved) and are still listed in
+     * the tab-bar overflow dropdown so the user can re-activate or unhide
+     * them. Defaults to `false` so existing persisted blobs deserialize as
+     * fully visible. Toggled by [WindowCommand.SetTabHidden].
+     *
+     * @see WindowCommand.SetTabHidden
+     */
+    val hidden: Boolean = false,
 )
 
 /**
--- a/server/src/main/kotlin/se/soderbjorn/termtastic/Application.kt
+++ b/server/src/main/kotlin/se/soderbjorn/termtastic/Application.kt
@@ -872,6 +872,7 @@ private suspend fun handleWindowCommand(text: String, ctx: WindowConnectionConte
         is WindowCommand.AddTab -> WindowState.addTab()
         is WindowCommand.CloseTab -> WindowState.closeTab(cmd.tabId)
         is WindowCommand.RenameTab -> WindowState.renameTab(cmd.tabId, cmd.title)
+        is WindowCommand.SetTabHidden -> WindowState.setTabHidden(cmd.tabId, cmd.hidden)
         is WindowCommand.MoveTab -> WindowState.moveTab(cmd.tabId, cmd.targetTabId, cmd.before)
         is WindowCommand.AddPaneToTab -> {
             // Inherit cwd from the anchor pane so a new terminal split off
--- a/server/src/main/kotlin/se/soderbjorn/termtastic/WindowState.kt
+++ b/server/src/main/kotlin/se/soderbjorn/termtastic/WindowState.kt
@@ -399,6 +399,60 @@ object WindowState {
     }
 
     /**
+     * Toggle [TabConfig.hidden] on [tabId]. Hidden tabs are dropped from the
+     * tab strip rendering on the client and surfaced only via the tab-bar
+     * overflow dropdown — the rest of the tab (panes, focus, PTY sessions)
+     * survives untouched.
+     *
+     * Policy:
+     *  - Refuse if the tab id is unknown or the flag is already in the
+     *    requested state (no-op snapshot churn).
+     *  - Refuse if hiding [tabId] would leave zero visible tabs — the strip
+     *    must always have at least one button so the user can navigate.
+     *  - When hiding the currently active tab, switch the active selection
+     *    to the next visible neighbour (preferring the tab that takes the
+     *    same visual slot, mirroring [closeTab]'s neighbour fallback).
+     *
+     * Called from `handleWindowCommand` for [WindowCommand.SetTabHidden].
+     *
+     * @param tabId the id of the tab to hide or unhide
+     * @param hidden `true` to hide, `false` to unhide
+     * @see TabConfig.hidden
+     * @see WindowCommand.SetTabHidden
+     */
+    fun setTabHidden(tabId: String, hidden: Boolean) = synchronized(this) {
+        val cfg = _config.value
+        val idx = cfg.tabs.indexOfFirst { it.id == tabId }
+        if (idx < 0) return@synchronized
+        val tab = cfg.tabs[idx]
+        if (tab.hidden == hidden) return@synchronized
+        // Refuse to hide the last visible tab — the strip must always show
+        // at least one button.
+        if (hidden) {
+            val visibleAfter = cfg.tabs.count { !it.hidden && it.id != tabId }
+            if (visibleAfter == 0) return@synchronized
+        }
+        val newTabs = cfg.tabs.toMutableList()
+        newTabs[idx] = tab.copy(hidden = hidden)
+        // If we just hid the active tab, hop to the next visible neighbour
+        // so the strip still highlights something. Same fallback shape as
+        // closeTab: pick the visible tab whose new index is closest to the
+        // hidden one's old index, biased toward the tab that took its slot.
+        val newActive = if (hidden && cfg.activeTabId == tabId) {
+            val visibleIds = newTabs.filter { !it.hidden }.map { it.id }
+            // Walk forward from the hidden tab's old slot, then backward,
+            // looking for the first visible tab in either direction.
+            val forward = (idx until newTabs.size).firstNotNullOfOrNull { i ->
+                newTabs[i].takeIf { !it.hidden }?.id
+            }
+            forward ?: visibleIds.lastOrNull()
+        } else {
+            cfg.activeTabId
+        }
+        _config.value = cfg.copy(tabs = newTabs, activeTabId = newActive)
+    }
+
+    /**
      * Set the display title of [tabId] to [title] (trimmed, max 80 chars).
      * No-op if the title is empty or unchanged.
      *
--- a/web/src/jsMain/kotlin/se/soderbjorn/termtastic/OverflowMenu.kt
+++ b/web/src/jsMain/kotlin/se/soderbjorn/termtastic/OverflowMenu.kt
@@ -1,0 +1,309 @@
+/**
+ * Tab-bar overflow menu for the Termtastic web frontend.
+ *
+ * Anchored at the far-right edge of the tab strip via a static button in
+ * `index.html` (`#tab-overflow-button`). Opens a dropdown containing:
+ *
+ *  1. **Rename** — inline-rename the active tab. Falls back to a
+ *     `window.prompt` when the active tab is hidden (no strip button for
+ *     its label to become an input).
+ *  2. **Close** — close the active tab. Disabled when only one tab exists,
+ *     so the user can't end up with zero tabs. Triggers the same
+ *     confirmation dialog the old per-tab menu used.
+ *  3. **Hide / Unhide** — toggle [TabConfig.hidden] on the active tab.
+ *     Disabled when there is only one visible tab left (the strip must
+ *     always show something). Hiding the active tab makes the server
+ *     switch the active selection to the next visible neighbour — see
+ *     [WindowState.setTabHidden] for the server-side rule.
+ *  4. **A separator**, followed by one entry per hidden tab. Each entry
+ *     shows the content-kind icon of the tab's first pane (via
+ *     [iconSvgForLeafKind]) so the user can tell a file-browser tab from
+ *     a terminal tab at a glance. Clicking an entry sends
+ *     [WindowCommand.SetActiveTab] — the tab becomes active but stays
+ *     hidden; the user chooses whether to unhide afterwards.
+ *
+ * This file replaces the per-tab hover-menu that used to open next to each
+ * tab label. The overflow menu is a single, always-reachable affordance
+ * that doesn't steal horizontal space from the tab strip the way the old
+ * hover icon did.
+ *
+ * @see WindowCommand.SetTabHidden for the hide/unhide command
+ * @see WindowCommand.SetActiveTab for the activate-hidden-tab path
+ * @see iconSvgForLeafKind for the per-kind icon used in the hidden list
+ * @see renderConfig the tab-strip renderer that skips `hidden == true` tabs
+ */
+package se.soderbjorn.termtastic
+
+import kotlinx.browser.document
+import kotlinx.browser.window
+import org.w3c.dom.HTMLElement
+
+/** Pencil glyph for the "Rename tab" row. Stroke-only so it inherits the
+ *  menu row's `currentColor`. Sized to match the 14×14 leaf icons. */
+private val ICON_OVERFLOW_PENCIL =
+    """<svg viewBox="0 0 16 16" width="14" height="14" fill="none" stroke="currentColor" stroke-width="1.4" stroke-linecap="round" stroke-linejoin="round"><path d="M2.5 13.5l1-3 7-7a1.4 1.4 0 0 1 2 2l-7 7-3 1z"/><line x1="9.5" y1="4.5" x2="11.5" y2="6.5"/></svg>"""
+
+/** Close (X) glyph for the "Close tab" row. Mirrors the pane-header close
+ *  icon style at 14×14 instead of 16×16 so it fits inline with the leaf-
+ *  kind icons used for hidden tabs. */
+private val ICON_OVERFLOW_CLOSE =
+    """<svg viewBox="0 0 16 16" width="14" height="14" fill="none" stroke="currentColor" stroke-width="1.4" stroke-linecap="round" stroke-linejoin="round"><line x1="4" y1="4" x2="12" y2="12"/><line x1="12" y1="4" x2="4" y2="12"/></svg>"""
+
+/** Eye-with-slash glyph for the "Hide tab" row — same eye outline as
+ *  [ICON_OVERFLOW_EYE] with a diagonal strike-through. */
+private val ICON_OVERFLOW_EYE_OFF =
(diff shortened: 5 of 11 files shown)


soderbjorn:
**Claude Code** (an AI coding agent) picked up this PR via the `/pick-review` skill and reviewed the changes. This is not a human review — treat it as a second set of eyes, not an approval. Note up front: this is one of two independent PRs that close issue #2 — PR #7 is the other attempt and has already received its own review.

## Summary of changes
Adds a `hidden: Boolean` flag to `TabConfig` plus a `WindowCommand.SetTabHidden` command and a matching `WindowState.setTabHidden` mutator. Replaces the per-tab hover `.tab-menu` affordance with a single "⋯" overflow dropdown anchored between the scrolling tab bar and the header-actions cluster via a static `#tab-overflow-dropdown` element in `index.html`. The dropdown offers Rename / Close / Hide·Unhide for the active tab, followed by a labelled "Hidden tabs" section listing each hidden tab with an icon pulled from its first pane's content kind. Hidden tabs are dropped from the strip render loop but still get their content `<section>`, so an active+hidden tab keeps displaying. The shared leaf-icon SVG palette is extracted from `Sidebar.kt` into a reusable `iconSvgForLeafKind(contentKind, isLink)` helper used by both the sidebar and the overflow menu.

Server-side policy: refuses to hide the last visible tab, and hops `activeTabId` to the next visible neighbour when the active tab is hidden — the PR body calls this "same fallback shape as `closeTab`".

## Functional review
The happy path matches issue #2 faithfully: single far-right ⋯ button, the three system actions with icons, a separator, and per-hidden-tab rows with content-kind icons. `Hide` / `Unhide` labels flip correctly, and the client disables both "Hide when only one visible tab remains" and "Close when only one tab exists" for parity with the server-side refusals — good UX (`web/.../OverflowMenu.kt:287`, `:307`).

A handful of concrete edge cases worth walking through:

- **Close no longer plays the leaving animation.** The old per-tab menu's Close item used `tabWrap.classList.add("leaving"); window.setTimeout({ launchCmd(CloseTab) }, 220)` — a 220 ms exit animation. `OverflowMenu.kt:296` dispatches `CloseTab` synchronously without touching `.leaving`. The tab now vanishes instantly. That's a minor UX regression vs. what `main` does today (and, incidentally, PR #7 kept the animation for its Close item). A one-liner to restore: look up the `.tab-button[data-tab=...]` element in the close handler, add `.leaving`, and delay the `launchCmd` by 220 ms.
- **Rename via `window.prompt` on a hidden active tab** (`OverflowMenu.kt:271-279`). Functionally it works, but `window.prompt` in Electron renderer processes is known to behave oddly or be disabled outright in some configurations. Worth a quick sanity check in the actual Electron build — the web-only build will obviously show the native browser prompt, but the Electron shell may produce a deprecation warning or a no-op depending on how the BrowserWindow is configured.
- **Clicking a hidden tab activates it without unhiding** (`OverflowMenu.kt:363`). Intentional per the PR body. The issue is silent on this; both design interpretations are defensible, and the tooltip `Activate '…' (stays hidden)` on each row signals the behaviour well.
- **Active+hidden UX.** Same state as PR #7: content showing with no `.tab-button.active` on the strip. `positionActiveIndicator` already early-returns on missing active (`web/.../TabBar.kt:29`), so the indicator rail just disappears. The ⋯ button is the only cue that Hide was toggled — a subtle visual on the overflow button itself (e.g. a small eye-off badge) when the active tab is hidden would reinforce the state, but this is a polish item, not a gap.
- **Keyboard.** The dropdown has no arrow-key navigation or type-to-filter; users must click. Consistent with `pane-menu` and `layout-dropdown` in the existing code, so not a new regression, just a known ceiling.
- **Esc handler scope** (`OverflowMenu.kt:223`). The document-level keydown closes the overflow dropdown on any Escape press, regardless of focus. If a terminal has focus and consumes Escape, the terminal's handler still runs first; this only affects the overflow menu state. Fine.
- **Confirm dialog's HTML interpolation** (`OverflowMenu.kt:291-295`). The title is dropped into `<strong>${tabTitle}</strong>` unescaped — same pre-existing pattern the old per-tab menu used, so it's not a regression, but a tab title like `` would still execute. Unlikely to matter in practice (titles come from the server) but worth noting when moving this dialog somewhere it might be escaped properly.

No manual browser verification was performed by this review either — same gap the PR body explicitly flags.

## Technical review
The `iconSvgForLeafKind` extraction (`web/.../Sidebar.kt:57-77`) is the right call — it's the single piece of this PR I'd keep regardless of which of #5 / #7 wins, because it gives two surfaces a single source of truth for the leaf-kind icon set. The function is pure, properly KDoc'd, and replaces exactly the inline if-else ladder that was in `renderSidebar`. Clean.

`WindowState.setTabHidden` (`server/.../WindowState.kt:417-451`) encodes the policy in one readable function: bail on unknown id or no-op, refuse last-visible, then hop active forward with a last-visible fallback. Two small observations:

- The KDoc claims the fallback "prefer[s] the tab that takes the same visual slot, mirroring `closeTab`'s neighbour fallback" (line 401). The implementation searches `(idx until newTabs.size)` forward only, then falls back to `visibleIds.lastOrNull()`. If hiding a tab at the far-right edge with no visible tab to its right, the fallback lands on the *leftmost-end* visible tab (via `lastOrNull` of a filtered list — note that's the last in the *visible* list, which is usually to the left of the hidden one). That isn't "same visual slot" — a backward walk from `idx` before the last-visible fallback would be closer to the documented behaviour. Not a bug, but the code and the KDoc don't quite match. Cheap fix: `visibleIds.lastOrNull()` → walk `(idx - 1 downTo 0)` for the last visible tab before the hidden one, then last-in-list as the final fallback.
- `visibleIds` is computed but only used to read its last element. A `newTabs.lastOrNull { !it.hidden }?.id` inline would drop the intermediate list allocation. Very minor.

Dead selectors in `main.kt:344` survive this PR: the global outside-click handler still queries `.pane-menu.open, .tab-menu.open, .tab-menu-list.open, .pane-split-flyout.open, .pane-flyout-wrap.open`. `.tab-menu.open` and `.tab-menu-list.open` no longer match anything once this PR lands — the classes are gone from the codebase. Leaving them as a no-op isn't harmful, but it's dead grep-bait; should be trimmed in this PR.

Opening the overflow dropdown doesn't close other open dropdowns. `initTabOverflowMenu`'s click handler (`OverflowMenu.kt:207-215`) calls `ev.stopPropagation()`, which cancels the global `main.kt` outside-click handler that closes `.pane-menu.open` and friends. The old per-tab menu deliberately iterated and closed peers (`WindowConnection.kt`'s deleted `run {}` block did so via `querySelectorAll(".tab-menu.open, .pane-menu.open")`). If you want the overflow button to dismiss any other open menu on open, add the same sweep in the `populateTabOverflowMenu` call path.

Accessibility of the disabled rows is half-done (`OverflowMenu.kt:377-380`): `aria-disabled="true"` is set but the native `disabled` attribute is not. Since each item is a `<button>`, setting both would be the standard combo — it also prevents keyboard focus landing on a non-actionable row. Trivial to add.

`buildActionItem`'s `enabled` parameter (`OverflowMenu.kt:370`) is a clean seam; same for `emptyOverflowItem`. The `populateTabOverflowMenu` function is doing a fair amount of work (roughly 130 lines), but the sections are visually separated with clear comments and the helpers pull the right amount of glue out. Readable.

CLAUDE.md compliance: the new file has a file-level doc block at line 1 (`OverflowMenu.kt:1-30`) listing every menu item's purpose and the relevant `@see` cross-references; every function (public and private) has a KDoc with purpose, caller context, `@param`/`@return` where applicable, and at least one `@see`. `TabConfig.hidden` and `WindowCommand.SetTabHidden` KDoc accurately describe behaviour — no stale-doc mismatch like the one in PR #7.

The `tab-overflow-menu` reuses the `.dropdown-menu` base class for positioning (`styles.css:780-783` tunes width and padding), which is consistent with how `#debug-dropdown` and the layout dropdown are styled — this keeps the overflow menu visually aligned with the other header dropdowns. Good reuse.

No tests added for `setTabHidden`. The PR body notes `WindowState` lacks test scaffolding, which is true, but `setTabHidden`'s three policy branches (no-op / refuse-last / hop-active) are a textbook candidate for a small unit test once scaffolding exists. Not a blocker, but the gap should be explicit.

## Consequences
- Wire-protocol additions (`WindowCommand.SetTabHidden`, `TabConfig.hidden`) are backward-compatible via kotlinx-serialization defaults. Existing persisted blobs deserialize with `hidden = false`. No migration required.
- CSS class rename/removal (`.tab-menu*` → dead) is complete; `.tab-menu-list` is also stripped from the `.app-header` `no-drag` whitelist (`styles.css:262-263`). The old classes don't reappear elsewhere.
- `PaneHeader.ICON_CLOSE` is untouched (unlike PR #7, which promoted it to `internal`). This PR uses a fresh 14×14 X-glyph inline in `OverflowMenu.kt:22`. No cross-module visibility change.
- `index.html` gains a `<div id="tab-overflow-dropdown">` element at the app-header level. If there is an Electron-side preload script or Playwright selector that queries `.app-header > .header-actions` by position or asserts child count, it will see one more sibling — unlikely to matter but worth grepping if such tests exist.
- Outside-click dismissal runs two document-level handlers (the existing one in `main.kt` and the new one in `initTabOverflowMenu`). Both are cheap; no measurable perf cost.

## Risks
- **Blocker — competing PR.** PR #7 closes the same issue with a different design (button inside tab-bar after `+`, `isHidden` field instead of `hidden`, different active-tab policy). Both cannot land. The repo owner needs to pick. On the three concrete design questions that differ, this PR's choices are the more conservative / safer ones: placing the button outside the scrolling region (so it can't scroll off-screen), refusing to hide the last visible tab (so the strip is never empty), and hopping the active selection on hide (so there is always a visible active indicator). If the "leave-active-alone" semantics of PR #7 are not actually wanted, this PR is the better starting point.
- **Should-fix — Close animation regression.** The 220 ms `.leaving` class is not applied by the overflow Close path. Restore it for parity with the current on-main behaviour.
- **Should-fix — KDoc / implementation drift on `setTabHidden`'s fallback.** The doc says "same visual slot", the code searches forward only and then jumps to the last visible tab. Either align the docs to the implementation or fix the walk to be bidirectional.
- **Should-fix — dead selectors in `main.kt:344`.** `.tab-menu.open, .tab-menu-list.open` no longer exist. Clean them up in this PR rather than leaving them to confuse the next reader.
- **Nit — opening the overflow doesn't dismiss other open dropdowns.** Pane menus and layout menu can co-exist with the overflow dropdown until an outside click occurs.
- **Nit — `aria-disabled` without `disabled`.** Disabled action rows are keyboard-focusable. Add the native `disabled` attribute or use `tabindex="-1"` to match.
- **Nit — `window.prompt` rename path.** Verify the Electron build still shows something usable; consider a small in-app dialog if the flow becomes more than an edge case.
- **Nit — no test for the hide policy.** Small but policy-bearing mutator; worth a test once `WindowState` gets any test scaffolding.

## Ideas & alternatives
- Consider adding an inline "Unhide" affordance (e.g. a small eye-off button on the right edge of each hidden-tab row) so restoring a hidden tab takes one click rather than "click row → open menu again → Unhide". PR body already lists this as a follow-up and it fits the spec naturally.
- A subtle visual on the overflow button itself when the active tab is hidden (badge, tint, or the tab's title in the tooltip) would reinforce the active+hidden state without adding a new on-screen element.
- The hidden-tabs section has a "Hidden tabs" header (`OverflowMenu.kt:329-331`) that the other sibling PR does not include — nice touch, makes the menu structure self-explanatory. Worth keeping even if this PR is otherwise rebased.
- `populateTabOverflowMenu` could extract the "Hidden tabs" block into its own helper for symmetry with `buildActionItem` — not required, just improves the skim of `populateTabOverflowMenu`.

## Verdict
Merge with small changes — provided the duplicate-PR situation is resolved in this PR's favour. The design choices here are the more defensive ones (no empty strip, button can't scroll off, active tab hops on hide), documentation is accurate and CLAUDE.md-compliant, and the `iconSvgForLeafKind` extraction is a real hygiene win. The concrete asks before merge are: restore the 220 ms close animation, fix the `setTabHidden` doc/impl mismatch (or the impl), trim the dead `.tab-menu*` selectors from `main.kt:344`, and add the native `disabled` attribute to grey-out rows. Everything else in Risks is nit-level. I could not run the Gradle build or launch the Electron shell in this environment, so the `window.prompt` Electron behaviour and the end-to-end UI flow remain unverified by this review.

PR #7, opened 2026-04-23
Revamp tab bar with overflow menu and hideable tabs

### LINKED ISSUE #2: Revamp the tab bar and allow hidden tabs

soderbjorn:
Closes #2

## Summary
Replaces each tab's hover-to-reveal "⋯" menu with a single overflow menu anchored at the far right of the tab bar. The new menu operates on the active tab (Rename / Close / Hide·Unhide) and surfaces any tabs the user has hidden from the strip, each listed with a content-kind icon. Hidden tabs keep their panes and PTY sessions intact — they're just skipped when rendering the tab-button row.

## Background
Issue #2 asked for a far-right three-dot button that opens a dropdown with Rename / Close / Hide-or-Unhide / a separator / the list of all hidden tabs. Hidden tabs should not appear on the tab strip; the hover-to-the-right-of-the-label menu that existed per-tab should go away along with the horizontal space it reserved. Hidden tabs in the dropdown should use type-dependent icons; system menu items should use appropriately picked icons, reusing existing ones where possible.

## Approach
Three layers change:

1. **Data model (`clientServer`)** — `TabConfig` gains `isHidden: Boolean = false`. The default keeps legacy persisted blobs round-tripping as fully visible. A new `WindowCommand.SetTabHidden(tabId, hidden)` joins the existing tab commands.
2. **Server (`server`)** — `WindowState.setTabHidden` toggles the flag and emits a new `WindowConfig`. The active tab is left alone on purpose: hiding an active tab keeps its content on-screen and the menu flips to "Unhide" on next open. `Application.handleWindowCommand` dispatches the new command.
3. **Web client (`web`)** — a new `TabBarMenu.kt` file builds the far-right button and dropdown. `WindowConnection.renderConfig` now (a) filters hidden tabs out of the tab-button loop while still rendering their tab-panes into the wrap, (b) removes the old per-tab menu block entirely, and (c) calls `appendTabBarOverflowMenu` after the `+` button. `isOnlyFocusChange` also treats `isHidden` as structural so a toggle triggers a re-render. CSS swaps `.tab-menu*` rules for `.tab-bar-menu*` rules; `.tab-button.renaming .tab-menu` is gone along with the per-tab menu visibility gates.

Clicking a hidden tab in the list calls `setActiveTab` so it becomes the visible content — it stays hidden from the strip, and the menu's next open shows "Unhide" for it.

### Hiding leaves `activeTabId` untouched
I initially wrote `setTabHidden` to shift the active tab to a visible neighbour when the active tab was hidden, and to refuse to hide the last visible tab. Switching to the "leave active alone" model produced a simpler, more literal reading of the spec: "Hide (or Unhide) depending on if the tab is already hidden or not" implies the same tab can flip between states without the user losing their place. It also naturally produces the "Unhide" menu state when the active tab itself is hidden. The trade-off is that the user can end up with zero visible tab buttons — but the overflow menu still works, and the tab pane content is still on screen, so nothing is actually unreachable.

### System menu operates on the active tab
The issue doesn't spell out which tab Rename / Close / Hide apply to. With a single global dropdown there's no better target than the active tab, and that's consistent with how the old per-tab menu worked once the user had clicked a tab to activate it. Rename is suppressed when the active tab is hidden (there's no label element in the strip to edit); Hide / Unhide flip based on `isHidden`; Close requires at least two tabs total (same `canCloseTab` guard as before).

### New `TabBarMenu.kt` file instead of inlining in `WindowConnection.kt`
`WindowConnection.renderConfig` is already the longest function in the frontend. Pulling the menu builder (plus its icon constants) into a separate file keeps `renderConfig` focused on the tab-strip and pane rendering pipeline. The package-private `appendTabBarOverflowMenu` is the single entry point; the helpers are file-private.

### Hidden-tab icon is picked from the tab's first pane
A tab can hold panes of mixed kinds, so there's no obviously correct single icon. I went with the first pane's `content.kind` discriminator because it matches the common single-pane case and is stable across re-renders. An empty tab (no panes) gets a generic window icon. Picking "most common kind" or similar heuristics would be more complex without clearly better results.

### `ICON_CLOSE` in `PaneHeader.kt` is now `internal`
The issue called out "reused from other parts of the app, if possible/appropriate" for the close glyph. Making the existing constant `internal` lets `TabBarMenu.kt` share the exact same SVG. Rename/Hide/Unhide and the four tab-type icons are new (not duplicated anywhere) and live next to the menu builder.

### Close animation preserved
The old per-tab close flow added `.leaving` to the `tab-button` and delayed the `CloseTab` dispatch by 220ms to let the CSS exit animation play. The overflow menu's Close item does the same, looking up the active tab's button by `data-tab` before adding the class.

## Assumptions
- Users expect `Hide` on the active tab to keep its content visible rather than silently re-point to a neighbour. If not, the server change is a one-liner — this is the only behavioural judgment that can't be sanity-checked from the issue alone.
- Tab-type icon from the first pane is "good enough" for the dropdown. A future enhancement could aggregate across panes.

- **Auto-shift active tab on Hide** — more forgiving when the user has many tabs, but produces asymmetric Hide/Unhide behaviour and makes "Unhide" on an active hidden tab impossible. Rejected.
- **Disallow hiding the last visible tab** — would leave no empty-strip state. Rejected because the overflow button stays available as the sole affordance in that case and restricting it hurts the "Group, don't restrict" spirit.
- **Right-click a tab for its own menu** — would preserve per-tab intent, but the issue explicitly calls for a single far-right button, so this was out of scope.

## Verification
- `./gradlew :web:compileKotlinJs :server:compileKotlin :clientServer:compileKotlinMetadata` — passes locally (existing pre-existing warnings only; no new errors or warnings introduced).
- `./gradlew build -x test` — passes locally end-to-end, including the JS browser distribution and server shadow jar.
- No manual browser verification was performed in this environment — the Gradle build covers compilation/typing/serialization wiring but not live UI behaviour. A reviewer should spot-check: (a) the ⋯ button appears at the far right of the tab strip; (b) Rename opens the inline editor on the active tab's label; (c) Close shows the confirm dialog and plays the shrink-out animation; (d) Hide removes the tab-button and the menu flips to "Unhide"; (e) clicking a hidden tab in the list activates it without unhiding; (f) drag-and-drop reordering still works on visible tabs.

## Files of note
- `clientServer/.../WindowConfig.kt` — new `isHidden` field on `TabConfig`.
- `clientServer/.../WindowCommand.kt` — new `SetTabHidden` command.
- `server/.../WindowState.kt` — new `setTabHidden` mutator.
- `server/.../Application.kt` — one-line dispatch for the new command.
- `web/.../TabBarMenu.kt` — **new file**; the overflow menu builder and its icons.
- `web/.../WindowConnection.kt` — tab-strip loop filters hidden tabs, per-tab menu block removed, `appendTabBarOverflowMenu` call added, `isOnlyFocusChange` now tracks `isHidden`.
- `web/.../RenameHandlers.kt` — dropped the dead `.tab-menu.open` cleanup that the per-tab menu needed.
- `web/.../PaneHeader.kt` — `ICON_CLOSE` promoted from `private` to `internal` for reuse.
- `web/.../main.kt` — outside-click dismiss selector updated to `.tab-bar-menu*`.
- `web/.../styles.css` — `.tab-menu*` rules removed, `.tab-bar-menu*` rules added.

## Follow-ups
- Manual browser verification as listed in Verification.
- A theme-level touch pass on the new dropdown could align the hover/separator colours with tab-section palette overrides more precisely than the current `--separator` / `--surface` fallback.


soderbjorn:

## Summary of changes
Adds an `isHidden` flag to `TabConfig`, a new `WindowCommand.SetTabHidden` with a matching `WindowState.setTabHidden` mutator, and a new `TabBarMenu.kt` that renders a single "⋯" overflow menu anchored to the end of the tab bar (right of `+`). The per-tab hover-activated `.tab-menu` is deleted wholesale. The overflow menu exposes Rename / Close / Hide·Unhide for the active tab plus a separator and one row per hidden tab. Hidden tabs are skipped in the tab-button render loop but their pane content still renders into the wrap, so a hidden tab that is still active continues to show on screen.

## Functional review
The happy path matches issue #2 faithfully: far-right ⋯ button, the four system items, the separator, the per-pane-kind icon on each hidden-tab row (pencil / X / eye / eye-slash for system actions, plus four tab-kind icons). `canCloseTab` guard and the 220 ms leaving-animation timeout are preserved from the old per-tab menu, so Close still shrinks the tab before it disappears.

A few edge cases worth walking through before merge, because the issue didn't spell them out:

- **Active tab becomes hidden.** The server deliberately leaves `activeTabId` untouched (`server/.../WindowState.kt:394-413`). The tab strip then renders with **no `.tab-button.active` element at all** — `positionActiveIndicator` gracefully early-returns when none is found (`web/.../TabBar.kt:29`), so the orange rail simply disappears. The pane content keeps showing. This is internally consistent but visually unusual: the user sees content with no indicator on the strip telling them *which* tab they're on. The overflow menu flips to "Unhide" on next open, which is the only cue.
- **Every tab is hidden.** Nothing prevents it — the strip can render as just `+ ⋯`. Deliberate per the PR body; worth a sanity check that the user discovers the overflow button as their only affordance.
- **Rename on a hidden active tab.** Intentionally suppressed (`TabBarMenu.kt:314` requires `!activeIsHidden`) because `startTabRename` mutates the strip label, which doesn't exist for a hidden tab. That means renaming a hidden tab requires unhiding first — mild friction, acceptable MVP.
- **Clicking a hidden tab in the list** activates it without unhiding. The comment inline says so; the issue is silent on this choice. Previewing a hidden tab without committing to restore it is reasonable.
- **Tab with a top-level split** (no leaf at `panes[0]`). `tabIcon` (`TabBarMenu.kt:205-215`) reads `panes[0].leaf?.content?.kind`; for a non-leaf top pane that resolves to `null` → terminal icon, regardless of the split's real contents. Not a bug, just an inaccuracy in the hidden-tab row icon for split tabs.

No manual browser verification was performed by this review either — same gap the PR body calls out.

## Technical review
Stale KDoc on `WindowCommand.SetTabHidden` (`clientServer/.../WindowCommand.kt:303-314`):

> When hiding the currently-active tab, the server picks the next visible neighbour (or falls back to the first visible tab) so the user always has something selected.

This describes the *rejected* alternative. The PR's "Decisions & reasoning / Hiding leaves `activeTabId` untouched" section explicitly chose the opposite policy, and `WindowState.setTabHidden` implements the untouched-active behaviour. The doc block reads like a leftover from an earlier draft — should be rewritten to match the actual policy before merge, otherwise the next reader will be misled.

The `isOnlyFocusChange` update (`web/.../WindowConnection.kt:181-183`) correctly treats `isHidden` as structural so a toggle triggers a full re-render. `positionActiveIndicator`'s early return on missing `.tab-button.active` is what keeps the active-hidden state from NPE-ing — worth being aware of as a load-bearing bit of existing code this PR now depends on.

Icon handling (`TabBarMenu.kt:170-189`) introduces four new inline SVGs for pane kinds (`ICON_TAB_TERMINAL`, `ICON_TAB_FILES`, `ICON_TAB_GIT`, `ICON_TAB_EMPTY`). `Sidebar.kt:129-137` already maps `contentKind → SVG` inline for the sidebar tree, using a different SVG set (16×16 viewBox vs this file's 24×24). They're conceptually the same mapping rendered at two sizes. A shared `iconSvgForLeafKind(kind, size)` helper would prevent the two surfaces from drifting when a new pane content kind is added. PR #5's sibling implementation (see Risks) does exactly this extraction — worth borrowing even if the rest of that PR is discarded.

Promoting `PaneHeader.ICON_CLOSE` from `private` to `internal` (`PaneHeader.kt:31`) is the minimum change needed to reuse the X glyph and is fine. Rename / Hide / Unhide are net-new glyphs; there's nothing in `PaneHeader.kt` to reuse for those.

Documentation follows CLAUDE.md rigorously — the new file has a file-level block at line 1, every public/internal symbol has purpose + caller context + `@param` + `@see` tags. `TabConfig.isHidden` and `SetTabHidden` KDoc (aside from the staleness noted above) are likewise thorough. The KDoc on `WindowState.setTabHidden` is correct and explicit about the "active left unchanged" contract.

No unit test added for `setTabHidden`. The PR body notes `WindowState` has no existing test scaffolding, so this is consistent with the codebase, but the mutator encodes a deliberate policy decision that a reader can easily break with a "this looks like `closeTab`, let me add the same fallback" patch. A single test (even one that just asserts "hiding the active tab preserves `activeTabId`") would pin the contract.

The `showConfirmDialog` call for Close (`TabBarMenu.kt:320-323`) interpolates `${activeTitle}` straight into the HTML body — matches the pre-existing pattern from `WindowConnection.kt:360` verbatim, so it's not regressing anything, but this path has the same tab-title-is-trusted-HTML exposure the old code had.

## Consequences
- Wire-protocol additions (`SetTabHidden` command, `TabConfig.isHidden`) are backward-compatible via kotlinx-serialization defaults — existing persisted blobs round-trip as `isHidden = false`. No migration.
- CSS class rename (`.tab-menu*` → `.tab-bar-menu*`) is complete in this diff and has no external consumers in the repo.
- The `.app-header .tab-menu-list` rule in `styles.css` that enabled `no-drag` on the dropdown when rendered under the Electron title bar is correctly renamed to `.tab-bar-menu-list` (`styles.css:263`). Verified no other `.tab-menu*` references survive.
- `PaneHeader.ICON_CLOSE` visibility widens from `private` to `internal`. Confined to the `web` module, no cross-module impact.
- Outside-click dismissal (`main.kt:344`) now targets `.tab-bar-menu.open, .tab-bar-menu-list.open` — correctly updated.

## Risks
- **Blocker — competing PR for the same issue.** PR #5 ("Tab-bar overflow dropdown with hide/unhide support", also `Closes #2`) is a second, independent implementation of the exact same feature on the `issue-2-tab-bar-overflow-menu` branch, opened ~20 minutes before this one. Only one can land. The two diverge on three real design questions:
  - **Active-tab policy when hiding:** this PR leaves `activeTabId` untouched; PR #5 hops to the next visible neighbour (and refuses to hide the last visible tab).
  - **Button placement:** this PR puts ⋯ inside `.tab-bar` after the `+` button; PR #5 places it outside the scrolling region, between the tab bar and `.header-actions`, so it can't scroll out of view when the user has many tabs.
  - **Shared icon helper:** PR #5 extracts `iconSvgForLeafKind` from `Sidebar.kt`; this PR adds parallel icon constants.

  The repo owner needs to pick. The button-placement question in particular is the most user-visible difference and the one I'd weigh most heavily — this PR's approach means the overflow button can scroll off-screen exactly when the user has the most tabs and most wants to hide some.
- **Should-fix — stale KDoc on `SetTabHidden`** (see Technical review). Documents the wrong policy.
- **Nit — active+hidden UX state.** Content on screen with no `.tab-button.active` indicator on the strip is a first-in-the-codebase state. May be disorienting; consider at minimum adding a subtle visual cue (e.g. the overflow button tinted, or the active tab's title rendered into `menuBtn`'s tooltip) so the user sees *which* hidden tab is currently active without opening the menu.
- **Nit — empty tab strip.** Deliberate, but worth confirming once in a live session that the `+ ⋯` layout still looks intentional rather than broken.
- **Nit — icon drift.** See Technical review; shared helper would de-risk future additions.
- **Nit — no test for the policy.** `setTabHidden`'s "leave active alone" contract is easy to accidentally regress.

## Ideas & alternatives
- If PR #5's design is discarded in favour of this one, still pull the `iconSvgForLeafKind` extraction from it — it's the one piece of that PR that's pure technical hygiene regardless of which design wins.
- Consider an inline Unhide button on each hidden-tab row so the user can restore a hidden tab to the strip in one click instead of "activate → open menu → Unhide" (three clicks). PR #5 lists this as a follow-up; it's equally applicable here.
- The "Tab options" tooltip on the ⋯ button (`TabBarMenu.kt:298`) is fine but generic. If you end up with the active-hidden state feedback, the tooltip is a natural place to surface which tab is active ("Tab options — active: _title_").
- A tiny refactor in `TabBarMenu.kt`: the `menuWrap.classList.remove("open"); menuList.classList.remove("open")` pair appears on every item's click handler. Extracting a `closeMenu()` local would tighten the builder.

## Verdict
Not ready to merge as-is, primarily because of the duplicate-PR situation — this needs to be reconciled with PR #5 before either can land. If this design wins, follow-ups should be: fix the stale `SetTabHidden` KDoc, borrow the shared icon helper from PR #5, and decide whether the active-hidden UX state needs a visual cue. The implementation itself is well-structured, documentation-compliant, and faithful to the issue where the issue is specific — the main concerns are a legitimately competing alternative and one demonstrable doc/code mismatch. I could not run the Gradle build or exercise the UI in this environment, so "does it actually render correctly" remains unverified here.

PR #10, opened 2026-04-23
Fix unclickable new-tab (+) button in Electron (closes #9)

### LINKED ISSUE #9: Broken new tab functionality
I am unable to click on the + sign in the tab bar to create a new tab. Nothing happens when I click or hover on it.

soderbjorn:
Closes #9

## Summary
The `+` button in the tab bar did nothing when clicked (and showed no hover feedback) in the Electron app. Cause: the button lives inside `.app-header`, which is a `-webkit-app-region: drag` region, and it was missing from the `no-drag` opt-out CSS rule that every other interactive header element already has. Adding `.tab-add` to that rule restores normal click and hover behaviour.

## Background
Issue #9 reports: *"I am unable to click on the + sign in the tab bar to create a new tab. Nothing happens when I click or hover on it."*

The Electron browser windows use `titleBarStyle: "hiddenInset"`, so the native title bar is gone. To keep the window draggable, `styles.css` marks the entire `.app-header` as `-webkit-app-region: drag`. That turns the header into an OS-level drag surface, which means the OS captures pointer events on it (clicks and hover effects are swallowed unless a child opts back out).

The existing rule at `web/src/jsMain/resources/styles.css:259` opts a hand-picked set of children back into normal interactivity:

```
.app-header .icon-button,
.app-header .tab-button,
.app-header .dropdown,
.app-header .dropdown-menu,
.app-header .tab-menu-list { -webkit-app-region: no-drag; app-region: no-drag; }
```

`.tab-add` — the `<button class="tab-add">+</button>` element created dynamically in `WindowConnection.kt:397-401` and appended to `#tab-bar` inside `.app-header` — was simply missing from that list. In a plain browser (`-webkit-app-region` is a no-op there) the button worked fine, which is presumably why this wasn't caught earlier. In Electron it's been silently broken.

## Approach
One-line CSS change: add `.app-header .tab-add` to the `no-drag` selector list. The existing comment block above the rule already documents why this list exists, so no new explanation was needed in the stylesheet.

### Opt `.tab-add` into the existing `no-drag` list rather than set `no-drag` inline on the button
`WindowConnection.kt` already relies on CSS class matching to grant `no-drag` to every other header button, and the opt-out list is colocated with the `.app-header { drag }` rule that creates the problem in the first place. Keeping the decision in CSS keeps all native-chrome opt-ins in one place — if someone later tweaks the drag policy they'll find every opt-out in the same rule. Setting the style inline on the DOM-built button would scatter the policy.

### Scope kept to `.tab-add` only
I audited every other element that currently lives inside `.app-header`:

- `.icon-button` (sidebar toggle, new-window, appearance, theme, settings, about, debug): already listed.
- `.tab-button` (each tab): already listed.
- `.tab-menu` / `.tab-menu-button` (per-tab hover menu): these are descendants of `.tab-button`, and `-webkit-app-region` inherits, so `no-drag` on the ancestor carries down.
- `.tab-menu-list`: already listed (it's appended to `document.body`, so the rule is technically a no-op at that location, but harmless).
- `.tab-active-indicator`: has `pointer-events: none`, not an interactive element — drag behaviour is irrelevant.
- `.dropdown`, `.dropdown-menu`: already listed.

`.tab-add` is the only missing case. A wider refactor (e.g. a `.app-header-interactive` umbrella class, or inverting the rule to tag the header background as `drag` instead of the whole header) would be cleaner in principle but out of scope for a bug fix — and would risk introducing its own regressions in Electron drag behaviour.

## Assumptions
- The user reported the bug while running the Electron build. A plain browser is unaffected by `-webkit-app-region`, so the diagnosis only explains the Electron failure mode. If the bug also reproduces in a plain browser, there is a separate cause that this PR does not fix — but the issue text (no click *and* no hover) matches exactly the Electron drag-region symptom.

- **Move `.tab-add` out of `.app-header` into `.header-actions` or a sibling container.** Would also work but changes layout/CSS ordering and is a bigger blast radius for a one-line bug fix.
- **Set `addBtn.style.webkitAppRegion = "no-drag"` in `WindowConnection.kt`.** Works, but splits the native-chrome policy between CSS and Kotlin.
- **Invert the whole policy (default no-drag, tag only a blank drag strip at the top).** Correct long-term design but well outside the scope of #9, and risks changing feel of window dragging across both platforms.

## Verification
- `./gradlew :web:jsBrowserProductionWebpack` — **passes locally** (webpack compiles with 2 pre-existing size-limit warnings unrelated to this change).
- **Manual in-app verification: not performed.** I could not launch the Electron shell from this environment. The fix is a one-line CSS addition that mirrors the pattern already proven to work for `.icon-button`, `.tab-button`, and the dropdowns — but the repo owner should click `+` in an Electron build to confirm the symptom is gone before merging. Hover feedback (`.tab-add:hover` background/colour change) should also return alongside the click.

## Files of note
- `web/src/jsMain/resources/styles.css` — the one-line change: add `.app-header .tab-add` to the existing no-drag opt-out rule (around line 262).

## Follow-ups
- Consider inverting the drag-region policy (default no-drag, tag a dedicated blank strip as drag) so that future header elements don't have to remember to opt out. Out of scope for this fix.
- No unit/UI test is added. `-webkit-app-region` is Electron-only; asserting it in jsDOM-style tests would require mocking that the repo doesn't currently do. The policy is one CSS rule, easy to eyeball, so a targeted test feels like overkill vs. the ongoing maintenance cost.

### DIFF
--- a/web/src/jsMain/resources/styles.css
+++ b/web/src/jsMain/resources/styles.css
@@ -258,6 +258,7 @@ body.appearance-light *::-webkit-scrollbar-thumb:hover {
 }
 .app-header .icon-button,
 .app-header .tab-button,
+.app-header .tab-add,
 .app-header .dropdown,
 .app-header .dropdown-menu,
 .app-header .tab-menu-list {


soderbjorn:

## Summary of changes
One-line CSS addition: `.app-header .tab-add` joins the existing `no-drag` opt-out list at `web/src/jsMain/resources/styles.css:259-266`. Behaviourally, the `+` button in the tab bar becomes clickable and regains its hover feedback inside the Electron shell, where `titleBarStyle: "hiddenInset"` forces `.app-header` to double as the window-drag region. In a plain browser nothing changes (`-webkit-app-region` is a no-op there) — this fix is purely for the Electron build.

## Functional review
The diagnosis is correct. `.app-header` is declared `-webkit-app-region: drag` at `styles.css:255-258`, and the sibling rule at `:259-266` hand-picks which children opt back out. `.tab-add` is the dynamically-created `<button class="tab-add">+</button>` appended to `#tab-bar` at `WindowConnection.kt:398`, and it was simply missing from that list. Matching the existing pattern (`.icon-button`, `.tab-button`, `.dropdown`, `.dropdown-menu`, `.tab-menu-list`) is exactly the right shape of fix.

Spot-checked the rest of the interactive surface inside `.app-header` against `index.html:11-128` and `WindowConnection.kt`:

- Every static `.icon-button` (sidebar toggle, debug, layout, new-window, appearance, theme, settings, about) and every `.dropdown` wrapper is already covered.
- `.tab-button` instances are covered.
- `.tab-active-indicator` has `pointer-events: none` at `styles.css:373`, so drag-region handling doesn't matter for it.
- `.tab-menu` / `.tab-menu-button` (per-tab hover menu) are descendants of `.tab-button`. The PR body asserts `-webkit-app-region` inherits — worth flagging that the Chromium implementation of this property is technically non-inheriting (it's not formal CSS inheritance), but in practice Electron treats a `no-drag` descendant of a `no-drag` ancestor as also `no-drag`, so these currently work. Not a concern for this PR; just note it's an empirical behaviour rather than a spec guarantee if you ever need to lean on it.

I did not launch an Electron build to watch the `+` button respond, so "the user can now click it" is inferred from the CSS diff matching the existing working pattern, not observed. Same gap the PR body calls out.

## Technical review
The one-line addition is correctly placed (alphabetical-ish order relative to the existing list, though the list isn't strictly sorted anyway). No other files need to change — this is genuinely a single-line bug.

Minor stylistic consistency point: the existing selector list doesn't share a single scope tag, so adding `.app-header .tab-add` follows the established pattern. The alternative of collapsing all `.app-header .<class>` entries into `.app-header :is(.icon-button, .tab-button, …)` would be tidier but well outside the scope of a bug fix.

CLAUDE.md documentation rules don't apply here — stylesheet-only change with no Kotlin/JS symbols touched.

The PR body's "Decisions & reasoning" section is more thorough than the diff requires, but that's fine — the rationale for keeping the policy in CSS instead of inlining `-webkit-app-region: no-drag` on the Kotlin-constructed button is sound, and colocating the policy with the original drag declaration is the maintainable choice.

## Consequences
- **Electron-only behaviour change.** The `+` button becomes interactive. This is the intended fix; no user in a plain browser is affected.
- **Interaction with in-flight PRs #5 and #7 (tab-bar overflow dropdown).** Both of those PRs add a new interactive element under `.app-header`. PR #5 adds `<div id="tab-overflow-dropdown" class="tab-overflow-menu …">` between the scrolling tab bar and `.header-actions` — its root uses `.dropdown-menu`-style positioning but the wrapper class is new, so whichever of those PRs merges will need to add its own selector (or inherit the existing `.dropdown` coverage). PR #7 places the overflow button inside `.tab-bar` with class `.tab-bar-menu` / `.tab-bar-menu-button`, also a new name that isn't in the current list. Neither is this PR's problem to solve, but whoever merges the overflow PR should double-check the Electron click behaviour in the same way #9 surfaced the issue here — ideally before a user files #12 for the overflow button.
- No wire format, serialisation, schema, or settings migration impact.
- Build/CI impact: zero (CSS-only).

## Risks
- **None rising to blocker or should-fix level.** The change is a single selector addition that mirrors a proven pattern.
- **Nit — no regression test.** As the PR body acknowledges, `-webkit-app-region` is Electron-only and the repo has no jsDOM/Electron-level UI harness. A comment in the CSS (right above `.app-header`) already points at this failure mode; a reader-facing note about the class list being "must-update-when-adding-interactive-children" would reduce the chance of the same bug recurring, but that's a polish item.
- **Nit — drag-region policy inversion (mentioned by the author as a follow-up).** Inverting the policy so the default is `no-drag` and only a dedicated strip is `drag` would retire this whole class of bug. Right scope for a separate PR; noting it here so it doesn't get lost.

## Ideas & alternatives
- If you do invert the drag policy later (default no-drag, strip-only drag), the `hiddenInset` integration comment at `styles.css:239-254` is a natural place to rewrite the design note. Probably pairs well with the Electron title-bar work already hinted at in `main.kt:303-323`.
- As a lower-effort intermediate, consider `.app-header > *:not(.tab-active-indicator)` or a dedicated `.app-header-interactive` class on every interactive child. Each has its own trade-off (the `:not` selector is terser but breaks if someone adds another `pointer-events: none` overlay; the explicit class is more verbose but self-documenting). Neither is worth doing in this PR.

## Verdict
Merge as-is. The diagnosis is correct, the fix mirrors the established pattern, the blast radius is one selector, and the author's own verification gap (no live Electron pass) is clearly called out in the PR body — easy for the human reviewer to close by just clicking `+` once in the Electron build before merging. I could not launch the Electron shell from this environment, so I'm leaning on the CSS analysis rather than observed end-to-end behaviour; flagging as the one unverified leg.

PR #13, opened 2026-04-24
Add per-tab 'Hide in side bar' toggle (#12)

### LINKED ISSUE #12: Possibility to hide a tab from the side bar
The drop-down menu for a tab has a "Show in tabbar" / "Hide in tab bar" toggle. I would like a similar one called "Show in side bar" / "Hide in side bar", which should -- as you would expect -- hide it in the sidebar. This setting should be persisted in the server DB. And by the way, the other setting (tab bar toggle) should also be, if it's not already done.

soderbjorn:
Closes #12

## Summary
Adds a new "Show in side bar" / "Hide in side bar" entry to the tab-bar overflow (`⋯`) menu, mirroring the existing "Show/Hide in tab bar" affordance. Hiding a tab from the sidebar removes it from the left sidebar tree but leaves it present in the tab strip and keeps all its panes and PTY sessions alive. The setting is persisted server-side in SQLite alongside the rest of `WindowConfig`.

## Background
From issue #12:

> The drop-down menu for a tab has a "Show in tabbar" / "Hide in tab bar" toggle. I would like a similar one called "Show in side bar" / "Hide in side bar", which should -- as you would expect -- hide it in the sidebar. This setting should be persisted in the server DB. And by the way, the other setting (tab bar toggle) should also be, if it's not already done.

The existing `TabConfig.isHidden` flag (tab-bar visibility) is **already** persisted via `SettingsRepository.saveWindowConfig()` — it round-trips through the same JSON blob under the `window.config.v3` key. No migration was needed for the existing toggle; the new flag follows the same path.

## Approach

The change follows the same layered pattern as the existing tab-bar hide toggle, end to end:

1. **Data model (`clientServer/WindowConfig.kt`)** — a new `isHiddenFromSidebar: Boolean = false` property on `TabConfig`. The default ensures legacy persisted configs deserialize as fully visible.
2. **Command (`clientServer/WindowCommand.kt`)** — a new `WindowCommand.SetTabHiddenFromSidebar(tabId, hidden)` sealed-class variant, registered with `@SerialName("setTabHiddenFromSidebar")`.
3. **Server handler (`server/WindowState.kt`)** — `setTabHiddenFromSidebar()` mirrors `setTabHidden()`: idempotent copy-on-write update to the `_config` StateFlow, synchronized, no-op when the flag already matches. Persistence is automatic — the existing debounced `collectLatest` loop in `Application.kt` picks up every `_config` mutation and upserts the serialized blob into SQLite.
4. **Command dispatch (`server/Application.kt`)** — one new branch in the `handleWindowCommand` `when` block.
5. **Web UI — tab-bar menu (`web/TabBarMenu.kt`)** — a second "Hide/Show" row is appended to the active-tab section of the overflow menu, directly below the existing tab-bar toggle. Shares the same `ICON_HIDE` / `ICON_UNHIDE` SVGs and `menuRow` builder.
6. **Web UI — sidebar render (`web/Sidebar.kt`)** — `renderSidebar` now skips tabs whose `isHiddenFromSidebar` flag is `true` before building the header/children DOM.

### Separate flag rather than reusing `isHidden`
I considered reusing the existing `isHidden` property to drive both surfaces, but the issue explicitly asks for an independent toggle — a user who declutters the sidebar likely still wants the tab in the strip (the sidebar is where the project already pre-filters mentally, so it's the natural place to hide rarely-used tabs). A single shared flag would couple two orthogonal affordances. Two flags makes each surface independently togglable and the intent readable at a glance in the config blob.

### Plain `continue` in the sidebar loop, no "Unlisted tabs" surfacing
The tab-bar overflow menu has a dedicated "Unlisted tabs" section that surfaces tab-bar-hidden tabs so they remain reachable. I deliberately did **not** mirror that for the sidebar: if a tab is hidden from the sidebar it is still present in the tab strip, so the user has an obvious alternate route to reach it. Adding a second surfacing section to the sidebar would (a) defeat the "declutter the sidebar" user intent and (b) create a redundant second entry point. The tab bar stays authoritative for reachability; the sidebar is a nav-convenience view.

### Placement in the overflow menu
The new row sits immediately after the existing "Hide in tab bar" row inside the "Active tab" section. Keeping them adjacent makes the two affordances self-documenting (you read them in a group), and places the new row in the same visual band the user is already used to finding visibility controls in.

### Shared icons (`ICON_HIDE` / `ICON_UNHIDE`)
Both toggles use the eye / eye-with-slash SVGs that the existing tab-bar toggle already uses. A sidebar-specific icon would add visual noise without improving scannability — the label text already distinguishes "in tab bar" vs "in side bar". This matches the guidance in `CLAUDE.md` around augmenting existing patterns rather than inventing new ones.

### Backward compatibility
`isHiddenFromSidebar: Boolean = false` as a defaulted Kotlin serializer field means:
- Existing persisted `window.config.v3` blobs without the field deserialize as fully visible — no migration needed.
- Older clients connecting to a newer server harmlessly ignore the field when round-tripping the JSON (kotlinx-serialization skips unknown fields by default on the ignore path this project already uses).
- Newer clients connecting to an older server never receive the field and use the default — also safe.

No schema version bump is required.

## Assumptions
- The issue's note that the tab-bar toggle "should also be [persisted], if it's not already done" is satisfied by the existing implementation (I verified `SettingsRepository.saveWindowConfig()` already encodes the full `WindowConfig` including every `TabConfig.isHidden` into the SQLite `settings` row).
- The user is fine with the new row living directly under the existing tab-bar toggle in the same "Active tab" section rather than in a new subsection.
- No sidebar-side context-menu affordance is required for now — the issue scopes the request to mirroring the existing drop-down, which lives in the tab bar.

- **Reuse `isHidden` for both surfaces** — rejected, see decision above.
- **Add a sidebar-side right-click menu to toggle visibility from the sidebar itself** — out of scope; the issue explicitly scopes the ask to the tab-bar drop-down. Easy follow-up if the user wants it.
- **Localize via client localStorage** — rejected per project convention: settings live server-side in SQLite only; no localStorage (see the project's client-server split).
- **Separate typed DTO for "visibility settings"** — rejected per project convention favouring a generic JSON blob over typed DTOs.

## Verification
- `./gradlew assemble` passes cleanly end to end (`BUILD SUCCESSFUL in 1m 13s`, 438 tasks) — covers `clientServer` (common), server, web JS (production bundle), androidApp, and the iOS framework link steps.
- Manual runtime verification was not performed — I did not launch the server + web client in this environment. Reviewer should smoke-test: open the `⋯` menu, toggle "Hide in side bar" on the active tab, confirm the tab disappears from the sidebar tree but stays in the strip and all its panes remain usable; then toggle back and confirm the tab reappears. Restart the server and confirm the setting persists.
- No existing tests exercise this area (the project does not ship unit tests for `WindowState` mutations in the changed file).

## Files of note
Load-bearing:
- `clientServer/src/commonMain/kotlin/se/soderbjorn/termtastic/WindowConfig.kt` — new `TabConfig.isHiddenFromSidebar` field.
- `clientServer/src/commonMain/kotlin/se/soderbjorn/termtastic/WindowCommand.kt` — new `SetTabHiddenFromSidebar` command.
- `server/src/main/kotlin/se/soderbjorn/termtastic/WindowState.kt` — new `setTabHiddenFromSidebar()` handler.
- `web/src/jsMain/kotlin/se/soderbjorn/termtastic/Sidebar.kt` — sidebar filter.
- `web/src/jsMain/kotlin/se/soderbjorn/termtastic/TabBarMenu.kt` — new menu row + doc updates.

Mechanical:
- `server/src/main/kotlin/se/soderbjorn/termtastic/Application.kt` — one-line dispatch branch added.

## Follow-ups
- Optional: add a sidebar-side affordance (e.g. right-click on a sidebar tab header) to toggle visibility from the sidebar itself. Not asked for in this issue.
- Optional: if more visibility dimensions get added in the future (e.g. "hide in tab bar on mobile only"), consider collapsing them into a single visibility mask to keep `TabConfig` flat.

### DIFF
--- a/clientServer/src/commonMain/kotlin/se/soderbjorn/termtastic/WindowCommand.kt
+++ b/clientServer/src/commonMain/kotlin/se/soderbjorn/termtastic/WindowCommand.kt
@@ -320,6 +320,26 @@ sealed class WindowCommand {
     data class SetTabHidden(val tabId: String, val hidden: Boolean) : WindowCommand()
 
     /**
+     * Mark a tab as hidden or visible in the left sidebar's tab tree. Hidden-
+     * from-sidebar tabs keep all their panes and PTY sessions alive and stay
+     * reachable through the tab strip; the web client simply skips them when
+     * rendering the sidebar so the user can declutter it independently of the
+     * tab bar. Unlike [SetTabHidden] there is no overflow-menu surfacing on
+     * the sidebar itself — the tab bar remains the way to reach a sidebar-
+     * hidden tab.
+     *
+     * Dispatched from the same far-right "⋯" tab-bar overflow menu as
+     * [SetTabHidden], via the "Show in side bar" / "Hide in side bar" entry.
+     *
+     * @param tabId the id of the tab to hide or unhide in the sidebar
+     * @param hidden `true` to hide the tab from the sidebar, `false` to show it
+     * @see TabConfig.isHiddenFromSidebar
+     */
+    @Serializable
+    @SerialName("setTabHiddenFromSidebar")
+    data class SetTabHiddenFromSidebar(val tabId: String, val hidden: Boolean) : WindowCommand()
+
+    /**
      * Add a new terminal pane to an existing tab (creates a new PTY session).
      *
      * @param tabId the id of the tab to add the pane to
--- a/clientServer/src/commonMain/kotlin/se/soderbjorn/termtastic/WindowConfig.kt
+++ b/clientServer/src/commonMain/kotlin/se/soderbjorn/termtastic/WindowConfig.kt
@@ -73,6 +73,24 @@ data class TabConfig(
      * @see WindowCommand.SetTabHidden
      */
     val isHidden: Boolean = false,
+    /**
+     * Whether this tab is hidden from the left sidebar's tab tree. Hidden-
+     * from-sidebar tabs still exist with all their panes and PTY sessions
+     * intact and still participate in the tab bar — the web client merely
+     * skips them when rendering the sidebar tree so the user can declutter
+     * the sidebar independently of the tab strip. Toggled from the tab-bar
+     * overflow menu's "Show in side bar" / "Hide in side bar" entry via
+     * [WindowCommand.SetTabHiddenFromSidebar]. Defaults to `false` so
+     * legacy persisted configs round-trip as fully visible.
+     *
+     * This is orthogonal to [isHidden]: a tab can be shown in the strip but
+     * hidden from the sidebar, and vice versa. Unlike [isHidden] there is
+     * no "overflow menu" surfacing for sidebar-hidden tabs — the tab bar
+     * remains the authoritative place to reach them.
+     *
+     * @see WindowCommand.SetTabHiddenFromSidebar
+     */
+    val isHiddenFromSidebar: Boolean = false,
 )
 
 /**
--- a/server/src/main/kotlin/se/soderbjorn/termtastic/Application.kt
+++ b/server/src/main/kotlin/se/soderbjorn/termtastic/Application.kt
@@ -874,6 +874,7 @@ private suspend fun handleWindowCommand(text: String, ctx: WindowConnectionConte
         is WindowCommand.RenameTab -> WindowState.renameTab(cmd.tabId, cmd.title)
         is WindowCommand.MoveTab -> WindowState.moveTab(cmd.tabId, cmd.targetTabId, cmd.before)
         is WindowCommand.SetTabHidden -> WindowState.setTabHidden(cmd.tabId, cmd.hidden)
+        is WindowCommand.SetTabHiddenFromSidebar -> WindowState.setTabHiddenFromSidebar(cmd.tabId, cmd.hidden)
         is WindowCommand.AddPaneToTab -> {
             // Inherit cwd from the anchor pane so a new terminal split off
             // from an existing pane starts in the same directory instead of
--- a/server/src/main/kotlin/se/soderbjorn/termtastic/WindowState.kt
+++ b/server/src/main/kotlin/se/soderbjorn/termtastic/WindowState.kt
@@ -431,6 +431,37 @@ object WindowState {
     }
 
     /**
+     * Mark [tabId] as hidden or visible in the left sidebar's tab tree.
+     * Sidebar-hidden tabs keep every pane and PTY session intact and still
+     * participate in the tab bar; the web client simply omits them from the
+     * sidebar render. No-op if the id is unknown or the flag already matches
+     * the requested value.
+     *
+     * Orthogonal to [setTabHidden]: a tab can be shown in the strip but
+     * hidden from the sidebar, and vice versa. The active tab and focus
+     * state are left unchanged — sidebar-hiding is a pure UI affordance.
+     *
+     * Called from `handleWindowCommand` on
+     * [WindowCommand.SetTabHiddenFromSidebar], dispatched by the tab-bar
+     * overflow menu's "Show in side bar" / "Hide in side bar" entry.
+     *
+     * @param tabId the id of the tab to hide or unhide in the sidebar
+     * @param hidden `true` to hide from the sidebar, `false` to show it
+     * @see TabConfig.isHiddenFromSidebar
+     * @see WindowCommand.SetTabHiddenFromSidebar
+     */
+    fun setTabHiddenFromSidebar(tabId: String, hidden: Boolean) = synchronized(this) {
+        val cfg = _config.value
+        val idx = cfg.tabs.indexOfFirst { it.id == tabId }
+        if (idx < 0) return@synchronized
+        val tab = cfg.tabs[idx]
+        if (tab.isHiddenFromSidebar == hidden) return@synchronized
+        val newTabs = cfg.tabs.toMutableList()
+        newTabs[idx] = tab.copy(isHiddenFromSidebar = hidden)
+        _config.value = cfg.copy(tabs = newTabs)
+    }
+
+    /**
      * Set the display title of [tabId] to [title] (trimmed, max 80 chars).
      * No-op if the title is empty or unchanged.
      *
--- a/web/src/jsMain/kotlin/se/soderbjorn/termtastic/Sidebar.kt
+++ b/web/src/jsMain/kotlin/se/soderbjorn/termtastic/Sidebar.kt
@@ -61,6 +61,11 @@ fun renderSidebar(config: dynamic) {
     val tabsArr = config.tabs as? Array<dynamic> ?: return
 
     for (tab in tabsArr) {
+        // Respect the per-tab sidebar-visibility flag. Tabs flagged as
+        // sidebar-hidden still render in the tab bar and retain all their
+        // panes and PTY sessions — they are merely omitted from the sidebar
+        // tree so the user can declutter it independently of the strip.
+        if ((tab.isHiddenFromSidebar as? Boolean) == true) continue
         val tabId = tab.id as String
         val tabTitle = tab.title as String
         val isActiveTab = tabId == activeTabId
--- a/web/src/jsMain/kotlin/se/soderbjorn/termtastic/TabBarMenu.kt
+++ b/web/src/jsMain/kotlin/se/soderbjorn/termtastic/TabBarMenu.kt
@@ -17,7 +17,9 @@
  *
  * @see renderConfig — the single caller; invoked after the visible tab
  *   buttons and the `+` add button have been appended.
- * @see WindowCommand.SetTabHidden for the hide/unhide server round-trip.
+ * @see WindowCommand.SetTabHidden for the tab-bar hide/unhide round-trip.
+ * @see WindowCommand.SetTabHiddenFromSidebar for the sidebar hide/unhide
+ *   round-trip.
  * @see WindowCommand.CloseTab / WindowCommand.RenameTab / WindowCommand.SetActiveTab
  *   for the other actions dispatched from this menu.
  */
@@ -155,12 +157,17 @@ private fun menuSeparator(): HTMLElement {
  *  - "Rename"  — opens the inline rename UI on the active tab's label.
  *  - "Close"   — closes the active tab (only when there is more than one
  *    tab, matching `canCloseTab` elsewhere in the renderer).
- *  - "Hide" / "Unhide" — toggles [TabConfig.isHidden] on the active tab.
- *    A hidden active tab stays active (its content still renders in the
- *    wrap) so the user doesn't lose their place.
+ *  - "Hide" / "Unhide" in tab bar — toggles [TabConfig.isHidden] on the
+ *    active tab. A hidden active tab stays active (its content still
+ *    renders in the wrap) so the user doesn't lose their place.
+ *  - "Hide" / "Unhide" in side bar — toggles
+ *    [TabConfig.isHiddenFromSidebar] on the active tab, independently of
+ *    the tab-bar flag. Dispatched via
+ *    [WindowCommand.SetTabHiddenFromSidebar].
  *  - A separator and one row per hidden tab — clicking a hidden tab
  *    activates it; the user can then Unhide it if they want the strip
- *    entry back.
+ *    entry back. (Hidden-from-sidebar tabs remain reachable through the
+ *    tab strip and so have no separate overflow section.)
  *
  * Invoked once per render from [renderConfig], after the visible tab
  * buttons have been appended. The `New tab` row at the top of the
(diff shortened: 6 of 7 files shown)


soderbjorn:

## Summary of changes
Adds a second per-tab visibility flag, `TabConfig.isHiddenFromSidebar`, orthogonal to the existing `isHidden` (tab-bar) flag, and wires it through the full stack: a new `SetTabHiddenFromSidebar` command, a matching `WindowState.setTabHiddenFromSidebar` handler, a dispatch branch in `Application.handleWindowCommand`, a filter in the sidebar renderer, and a new "Hide/Show in side bar" row in the tab-bar overflow menu directly below the existing "Hide/Show in tab bar" row. Persistence is inherited from the existing `window.config.v3` JSON blob path. The behavioural description in the PR body matches the diff — this is exactly the mirror affordance that issue #12 asks for.

## Functional review
Happy path: toggle the new menu row, server flips `isHiddenFromSidebar`, sidebar skips the tab on the next render. Server-side mechanics look right — `setTabHiddenFromSidebar` is synchronized, idempotent, and uses the same copy-on-write StateFlow update as `setTabHidden`, so the existing debounced persistence loop picks it up for free.

There is, however, one behavioural gap I want to flag before anything else:

**The client may not re-render the sidebar when only `isHiddenFromSidebar` changes.** `WindowConnection.kt:170-187` defines `isOnlyFocusChange`, which short-circuits `renderConfig` (including `renderSidebar`) when `prev` and `next` differ "only in per-tab `focusedPaneId`". Its comparison currently checks `id`, `title`, `isHidden`, and `stringify(panes)` per tab — but not `isHiddenFromSidebar`. So the flow after toggling the new menu entry is:

1. Client sends `SetTabHiddenFromSidebar`.
2. Server flips the flag on `_config` and broadcasts the new config.
3. Client receives the config, `renderConfig` runs, `isOnlyFocusChange(prev, next)` returns `true` (id/title/isHidden/panes all unchanged), `renderConfig` sets `currentConfig = config` and returns early.
4. `renderSidebar` is never called — the hidden tab is still visible in the sidebar until something else triggers a real rebuild (tab switch, pane change, rename, etc.).

The PR author noted they didn't exercise the UI, which is how this slipped past: `./gradlew assemble` won't catch it.

Aside from that, the rest of the UX looks sensible:
- The row is adjacent to the tab-bar toggle in the "Active tab" section, which reads naturally.
- The icon reuse (`ICON_HIDE` / `ICON_UNHIDE`) is fine — the labels disambiguate, and inventing a second eye variant would be noise.
- Reachability is preserved: a sidebar-hidden tab stays in the tab strip, so the deliberate decision to not add a "sidebar-hidden tabs" surfacing section is defensible. If a user also tab-bar-hides the same tab, it remains reachable via the existing overflow-menu "Unlisted tabs" section (which keys off `isHidden` only, which is correct).
- Defaulted field on `TabConfig` means old persisted blobs deserialize as fully visible, no migration needed.

Edge cases I'd still want to eyeball at runtime once the above bug is fixed:
- Sidebar-hiding the **active** tab: the sidebar loses the active-tab highlight entirely, which is consistent with "don't list this tab" but may feel odd. Worth a smoke test.
- Sidebar-hiding the **only** tab: cosmetically leaves an empty sidebar tree; the user still has the tab strip. Acceptable but confirm.

## Technical review
Structurally the change is clean: it follows the existing `setTabHidden` pattern one-for-one and keeps the new flag orthogonal to the old, which is the right call per the PR body's reasoning. Specific notes:

- `clientServer/.../WindowConfig.kt:76-93` — field is added with a default of `false`, good for backward compatibility.
- `clientServer/.../WindowCommand.kt:321-339` — new sealed-class variant is `@Serializable` + `@SerialName("setTabHiddenFromSidebar")`, mirrors `SetTabHidden` exactly.
- `server/.../WindowState.kt:453-461` — mirror of `setTabHidden`; idempotent, synchronized, no-op on missing id or unchanged flag. Correct.
- `server/.../Application.kt:877` — one-line dispatch added in the `when`; nothing else needed because the existing `_config` debounce handles persistence.
- `web/.../Sidebar.kt:63-68` — filter-at-loop-top is the simplest place to put this. Fine.
- `web/.../TabBarMenu.kt:205-284` — reads well; the new row's wiring is a near copy of the tab-bar row and uses the same `menuRow` helper + `launchCmd` pathway. Minor nit: `activeIsHiddenFromSidebar` is computed unconditionally even when `activeTab == null`, which is safe because of the `as? Boolean` coercion but makes the code read slightly less defensively than `activeIsHidden`. Not worth changing.
- KDoc/JSDoc coverage is thorough and cross-referenced on both sides of the new flag, which meets the project's `CLAUDE.md` documentation rules. Good.

The spelling "side bar" (two words) in the user-visible labels matches issue #12's wording. Internally the identifiers use `isHiddenFromSidebar` / "sidebar" as one word, matching the rest of the codebase. That inconsistency is fine — it's user-facing vs. code-facing — but worth being aware of if future copy review wants to standardise.

No tests were added, consistent with the existing lack of coverage for `WindowState` mutations. Not a regression.

## Consequences
- Settings blob gains one new field under `window.config.v3.tabs[*].isHiddenFromSidebar`. Backward-compatible in both directions (default `false`, unknown-field tolerance documented in the PR body).
- No schema migration required.
- No API surface changes beyond the new command variant.
- Build/CI impact: zero — same modules, same targets.
- No interaction with in-flight work as far as I can see on `main` (no other tab/sidebar refactors touching `WindowConnection.kt` or `WindowState.kt` recently).

## Risks
- **Blocker — sidebar doesn't update after toggling the new flag.** See the functional review above. Fix: extend the per-tab comparison in `isOnlyFocusChange` at `web/src/jsMain/kotlin/se/soderbjorn/termtastic/WindowConnection.kt:183` to also compare `isHiddenFromSidebar`, e.g.:
  ```kotlin
  if ((pt.isHidden as? Boolean ?: false) != (nt.isHidden as? Boolean ?: false)) return false
  if ((pt.isHiddenFromSidebar as? Boolean ?: false) != (nt.isHiddenFromSidebar as? Boolean ?: false)) return false
  ```
  The adjacent comment (`Hidden-state toggles change what the tab strip renders … so they are *not* focus-only changes.`) should be extended to mention the sidebar flag for the same reason.
- **Should-fix — no runtime smoke test.** The PR body acknowledges this, but the bug above is exactly the class of thing that a 30-second manual test would have caught. Worth doing before merge even after the fix, to confirm no secondary layout oddities (active tab + sidebar-hidden, only tab + sidebar-hidden).
- **Nit — future flag sprawl.** If a third visibility dimension appears, `TabConfig` will start collecting boolean flags per surface. The PR body already anticipates this as a future follow-up; no action now.

## Ideas & alternatives
- Once the `isOnlyFocusChange` fix is in, consider factoring the "fields whose change forces a full rebuild" into a small list-of-pairs or a data-class comparison, so the next visibility flag added can't silently miss this guard. Not needed today, but the pattern is beginning to repeat.
- A sidebar-side context-menu affordance (right-click on a header → Hide) is listed as a follow-up in the PR body. Reasonable to leave out of this PR.

## Verdict
Merge with small changes. The change is well-scoped, follows the existing pattern cleanly, and documents itself properly — but the `isOnlyFocusChange` omission means the user-visible behaviour doesn't actually work end-to-end in the common "toggle and observe" case. A one-line fix to `WindowConnection.kt:183` and a manual smoke test should be enough to clear it. I did not run the build or exercise the UI in this review — that's an explicit gap, but the bug is identifiable statically from the diff.

soderbjorn:
Good PR review. Please fix the "Blocker — sidebar doesn't update after toggling the new flag", that seems important!

soderbjorn:
**Claude Code** (an AI coding agent) picked up this feedback via the `/pick-followup` skill and pushed 593a81a.

> Good PR review. Please fix the "Blocker — sidebar doesn't update after toggling the new flag", that seems important!

Extended `isOnlyFocusChange` in `web/src/jsMain/kotlin/se/soderbjorn/termtastic/WindowConnection.kt:170-188` to also compare `isHiddenFromSidebar` per tab — exactly the one-line fix flagged in the review. Without this, the early-return short-circuit was setting `currentConfig` and skipping `renderSidebar`, so toggling "Hide in side bar" left the tab visible in the sidebar until an unrelated rebuild fired. Updated the adjacent comment to mention the sidebar flag for the same reason as the tab-bar flag. No other files touched — this is the minimal fix the reviewer recommended.

**Verification.** `./gradlew assemble` passes locally (`BUILD SUCCESSFUL in 26s`, 439 tasks). I did not exercise the toggle in a running browser session in this environment, so the runtime smoke test the reviewer recommended (toggle on, sidebar drops the tab; toggle off, tab returns; restart the server, setting persists) is still worth doing before merge.

🤖 This comment was posted by [Claude Code](https://claude.com/claude-code) acting autonomously — addressed in commit 593a81a. The human repo owner has not yet confirmed the change.

soderbjorn:

## Summary of changes
This is a re-review after commit 593a81a, which addresses the blocker flagged in the earlier review: the client's `isOnlyFocusChange` short-circuit ignored the new flag, so toggling "Hide in side bar" wouldn't cause the sidebar to rebuild until something else forced a full render. The fix extends the per-tab comparison in `web/src/jsMain/kotlin/se/soderbjorn/termtastic/WindowConnection.kt:184-185` to include `isHiddenFromSidebar`, with a matching update to the adjacent comment. Three lines added, nothing else changed.

## Functional review
The fix matches exactly what the previous review recommended and is structurally identical to the existing `isHidden` guard one line above — same `as? Boolean ?: false` coercion, same early-return-on-mismatch pattern. After this change the flow is:

1. Client sends `SetTabHiddenFromSidebar`.
2. Server flips the flag on `_config` and broadcasts.
3. Client receives the new config, `isOnlyFocusChange` now returns `false` because the per-tab flag differs, `renderConfig` proceeds through the full rebuild path, and `renderSidebar` re-runs at the end.
4. The sidebar loop in `web/src/jsMain/kotlin/se/soderbjorn/termtastic/Sidebar.kt:66` skips the sidebar-hidden tab, so it disappears from the tree.

The guard is correctly placed — inside the per-tab loop, before the `stringify(panes)` comparison — so it short-circuits cheaply and can't be bypassed.

The two edge cases I flagged earlier are still worth a 30-second runtime smoke test but appear fine on static inspection:
- **Sidebar-hiding the active tab.** `Sidebar.kt` only applies `active-tab` when the tab actually renders. A sidebar-hidden active tab simply produces no sidebar entry at all (no "orphaned" active-tab highlight). The tab remains active and usable through the tab strip and terminal wrap. Acceptable behaviour.
- **Sidebar-hiding the only tab.** Loop produces an empty `div.sidebar-tree`; the sidebar visually becomes empty. Cosmetic, not broken — the user still has the tab strip as the sole navigation path. Acceptable.

One additional sanity check worth noting: the "Unlisted tabs" section at the bottom of the tab-bar overflow menu filters by `isHidden` only (`TabBarMenu.kt:208`), so a tab that is both `isHidden` and `isHiddenFromSidebar` is still surfaced there — reachability is preserved. Good.

## Technical review
Minimal, surgical fix. The adjacent comment update ("The sidebar-visibility flag likewise changes what the sidebar tree renders, so toggling it must trigger a full rebuild.") keeps the intent documented for the next person who edits this function, which is exactly the right level of care here. No tests were added, which matches the existing lack of coverage for `renderConfig` / `isOnlyFocusChange` — not a regression.

One small forward-looking note (not blocking): `isOnlyFocusChange` now compares four per-tab fields explicitly (`id`, `title`, `isHidden`, `isHiddenFromSidebar`) plus a stringified `panes` snapshot. Each new visibility dimension added in future will require another explicit guard here, with the same failure mode if it's forgotten. If a third such flag ever lands, extracting the "fields whose change forces a full rebuild" into a small list (or comparing a hash/stringification of a per-tab projection) would make the guard harder to silently miss. Today, two explicit comparisons is still clearer than the abstraction would be — leave it as is.

Documentation on both new API symbols (`WindowCommand.SetTabHiddenFromSidebar`, `TabConfig.isHiddenFromSidebar`, `WindowState.setTabHiddenFromSidebar`) remains thorough and cross-referenced, satisfying the project's `CLAUDE.md` rules.

## Consequences
Unchanged from the original review: one new optional field in the persisted `window.config.v3` JSON blob (`tabs[*].isHiddenFromSidebar`, defaults to `false`), backward-compatible in both directions, no schema migration required, no API surface changes beyond the already-added command variant. Build/CI impact is zero. The PR author confirms `./gradlew assemble` passes locally (`BUILD SUCCESSFUL in 26s`, 439 tasks) after the fix.

## Risks
- **Blocker from the prior review is resolved.** No new blockers identified.
- **Should-fix — runtime smoke test still not performed.** Author acknowledges this explicitly. Given the earlier blocker was exactly the class of bug `./gradlew assemble` cannot catch, I'd want eyes on a browser session before merge: toggle "Hide in side bar" on the active tab, confirm it drops from the sidebar on the same render cycle (without a tab switch, rename, or pane change to force a rebuild); toggle back, confirm it returns; restart the server, confirm persistence; and eyeball the active-tab-only and last-tab edge cases. All three take under a minute together.
- **Nit — none.**

## Ideas & alternatives
None beyond the forward-looking note above about how a third visibility flag would interact with `isOnlyFocusChange`. Not actionable today.

## Verdict
Merge after a quick runtime smoke test. The fix is exactly the one-line change recommended, correctly placed, correctly commented, and structurally consistent with the surrounding guards. The only remaining gap is that the UI hasn't been exercised end-to-end in a live browser — which is the same gap the author flagged in the original PR body and which is how the earlier blocker slipped through. I did not run the build or exercise the UI in this re-review either; that is an explicit gap, but the fix is statically verifiable from the diff and the surrounding code.

Issue #15, opened 2026-04-24
Android app missing some terminal padding


soderbjorn:
Text appears sometimes slightly off the edge (or at least slightly too close to the edge) in the Android app while viewing a terminal. We should add some (minimal) padding.

<img width="1008" height="2244" alt="Image" src="https://github.com/user-attachments/assets/fdf23cd2-1a87-42b6-811b-1a4308e811ef" />

soderbjorn:
**Claude Code** (an AI coding agent) picked up this issue via the `/pick-issue` skill and opened #17 with a proposed fix.

The PR adds a small Compose-level inset (6 dp horizontal, 2 dp vertical) around the Termux `TerminalView` in `TerminalScreen.kt` so rendered glyphs no longer sit flush against the screen edges. The inset has to be applied on the hosting `Box` rather than as view-padding on the `TerminalView` itself, because Termux's vendored renderer ignores Android view padding and paints directly against `getWidth()` / `getHeight()`; shrinking the Compose box makes the view measure smaller and the grid adapts automatically.

🤖 This comment was posted by [Claude Code](https://claude.com/claude-code) acting autonomously — the repo owner has not yet reviewed the PR.

soderbjorn:
Imported into Lunicle as **LMX-11** (with full comment & attachment history).

Issue #1, opened 2026-04-23
Theme and color scheme renames


soderbjorn:
Rename TerminalScheme to ColorScheme and ThemeConfigPreset to Theme throughout the codebase, since that matches better what we call them in the UI. Change both classes and instances, functions, etc as needed. Change the filenames if needed too. Even go ahead and change serialised names (on the wire, in configuration storage and on web sockets, if present) - losing these preferences is OK, but if changing them would cause us to be unable to launch the app afterwards (and retain our stored scrollback logs) then you should omit it.

soderbjorn:
**Claude Code** (an AI coding agent) picked up this issue via the `/pick-issue` skill and opened #11 with a proposed fix.

The PR renames `TerminalTheme` → `ColorScheme` and `ThemeConfigPreset` → `Theme` (plus file renames, top-level list vals, and the now-recursive `Theme.theme` field → `Theme.colorScheme`). Scope notes and one interpretation call — treating the issue's "TerminalScheme" as a typo for `TerminalTheme`, since no `TerminalScheme` class exists — are documented in the PR body. Full Gradle build passes (client, web, server, androidApp).

🤖 This comment was posted by [Claude Code](https://claude.com/claude-code) acting autonomously — the repo owner has not yet reviewed the PR.

soderbjorn:
Imported into Lunicle as **LMX-3** (with full comment history).

PR #16, opened 2026-04-24
Bring back the Termtastic logo (lower-right, pulsating dot)

### LINKED ISSUE #14: Bring back the logo
A long time ago we used to have the label Termtastic in a nice font in the upper left corner, along with a green dot to the left of it. At different points throughout the project history, the green dot could also assume other colors depending on status. Please refer to the commit log for more details.

I want to bring back the logo but place it in a different place. It looks like the lower right corner would be a good fit - we have enough padding already, and also a bottom bar. Instructions:

1. To provide enough space, the bottom bar should always show, even if the server does not have Claude Code usage polling enabled. If the sidebar has no content, it should still take up space. 
2. The label Termtastic and the familiar green dot should be displayed in the lower right corner of the window, with a small amount of padding to the window corner. It should be painted on top of everything else on the window regardless of content underneath, which gives us more space to work with. The size of the label and dot should be chosen so that it should be a good fit and not overwrite the content inside the main tab content area. The green dot should be on the right side of the label, not like before.
3. The dot should pulsate in blue if any tab has ongoing work in progress, and in red if any tab has a window waiting for input, just like in the past. 

Image

soderbjorn:
Closes #14

## Summary
Reintroduces the Termtastic wordmark + status dot as a fixed overlay in the lower-right corner of the web UI. The dot mirrors the per-tab spinner/warning semantics at an aggregate level: it pulses blue when any tab is working, pulses red when any tab is waiting for input, and sits dim otherwise. The Claude usage bar is now always visible at the bottom of the window (empty when there's no data) so the logo has a consistent band of chrome to sit against.

## Background
Issue #14 asks for the old "Termtastic" label with its familiar coloured dot to come back, but repositioned to the lower-right corner (previously it lived upper-left in the header). The issue specifies three things:

1. The bottom bar should always be present, even when Claude Code usage polling is disabled — the bar should reserve space regardless of content.
2. The wordmark + dot should be overlaid in the lower-right corner, painted on top of everything, sized so they don't crowd the main content.
3. The dot should pulsate blue when any tab has work in progress and red when any tab is waiting for input (mirroring the historical behaviour, but now with colours tied to *work state* rather than socket state).

The issue text says "If the sidebar has no content, it should still take up space" — from context this is clearly a slip of the pen for "bottom bar." Proceeding with that interpretation since it's the only reading that fits the rest of the paragraph (the whole item is about the bottom bar) and makes the request internally consistent.

## Approach
Four moving parts:

1. **New overlay element** — `#app-logo` is appended directly inside `<body>` in `index.html` (outside `#app`), containing a `.app-logo-wordmark` span and a `.app-logo-dot` span. Being a sibling of `#app` rather than a child means it doesn't participate in flex layout and can overlay the whole window via `position: fixed; right; bottom; z-index: 100; pointer-events: none`.
2. **State aggregation** — a new `AppLogo.kt` module exposes `updateAppLogoState(sessionStates)`. The aggregation rule walks the session state map once: any "waiting" → red (wins, because it's action-required); else any "working" → blue; else idle. The function short-circuits on the first "waiting" it sees.
3. **Hook-up** — `updateStateIndicators()` in `WebState.kt` now calls `updateAppLogoState(sessionStates)` alongside the existing sidebar/tab-bar spinner updates. That function already runs on every state envelope and every config render, so the dot stays in sync without needing its own subscription.
4. **Always-on bottom bar** — `updateClaudeUsageBadge()` in `ClaudeUsageBar.kt` no longer hides the bar when `usage` is `null` or `undefined`; it empties its content and switches it to a new `.claude-usage-bar-empty` class instead. The base `.claude-usage-bar` rule gains a `min-height: 24px` so the bar reserves space whether or not it has content. The drag-divider above it still hides in the empty case (there's nothing to drag-collapse).

CSS delivers the dot itself: a 10 px circle coloured from `--text-secondary` in idle, with `@keyframes app-logo-pulse-blue` / `app-logo-pulse-red` animations added by the state modifier classes. The red pulse is shorter (1.2 s) and slightly scales the dot, matching the same "waiting is more urgent" cue `.pane-status-spinner.state-waiting` already uses via `fade-warning`.

### Overlay placed outside `#app` (sibling of `#app` inside `<body>`)
`#app` is a full-height flex column; adding the logo inside it would either push other content or force awkward absolute positioning inside a flex child. Putting the overlay at the same level as `#app` with `position: fixed` lets it float over the whole window cleanly and survives any future restructuring of the app's flex layout. The trade-off is that it lives slightly "outside" the Kotlin/JS rendering tree — but it's static markup that only needs one class update at a time, so that's fine.

### Dot colours: blue for working, red for waiting
The old design (per the issue) specifies the same blue/red mapping, so this is a direct carry-over. Blue being "normal progress" and red being "needs you" is also consistent with most chat-app conventions and with the existing `.pane-status-spinner.state-waiting` (which uses `--t-semantic-warn` yellow for the warning triangle — I kept the dot distinct with red so the two surfaces don't look identical; the in-bar triangle is informational, the overlay dot is a window-level alert).

### Dot on the right of the wordmark (not the left)
The issue explicitly calls this out as a change from the previous design. No judgement call here — just following the spec.

### Always-on bottom bar via `min-height` rather than a filler string
I considered rendering placeholder content ("no usage data") in the empty bar, but that would have felt noisy — the bar is most often empty for users who don't opt into Claude polling, and a permanent "no data" string becomes visual junk. A plain empty band at `min-height: 24px` reads as intentional chrome. When data arrives, `updateClaudeUsageBadge` fills it as before.

### Drag-divider stays hidden in the empty case
The usage-bar-divider's job is to let the user drag the usage bar shut. When the bar is empty there's nothing to shut — collapsing an already-empty band would only add an invisible hit target and a confusing interaction. Leaving the divider `display: none` keeps the behaviour well-defined: divider appears if and only if there's data to grab.

### Respect `prefers-reduced-motion`
The pulse animations are decorative; users who opt out of motion still get the colour change. A `@media (prefers-reduced-motion: reduce)` block disables the keyframes. This matches existing hygiene in the stylesheet.

### Logo font matches `.header-title` (Georgia serif) even though the header title itself is no longer rendered
The old design used Georgia for the wordmark; the `.header-title` CSS rule still exists from that era. Reusing that font family keeps the brand mark consistent with the (historical) header — and if the title is ever brought back, both will match without further work.

## Assumptions
- The issue text saying "the sidebar should still take up space" is a slip for "the bottom bar should still take up space." The paragraph is otherwise about the bottom bar. If that reading is wrong, sidebar behaviour would need separate work.
- Session state values of interest are the string literals `"working"` and `"waiting"`. Anything else (including `null`, `"done"`, or debug override labels other than these two) is treated as idle. This matches how `updateStateIndicators` already interprets the same map.
- `#app-logo` placed outside `#app` is acceptable architecturally. The repo has no existing precedent for body-level overlays, but `#app` being a layout container rather than a semantic root makes this reasonable.

- **Render the logo inside `.app-header`** — rejected: the issue explicitly wants it in the lower-right, and overlays inside the flex header would fight the header-drag Electron region and the header-collapse divider behaviour.
- **Embed the logo inside `.claude-usage-bar`** — rejected: when the bar has data, the logo would either overlap usage items or push them around. Fixed overlay avoids the layout coupling entirely.
- **Use the tab-bar spinner's yellow for the "waiting" dot** — rejected: a second yellow indicator in the same screen area would conflate the per-tab warning with the window-level alert. Red gives the overlay its own voice.
- **Show a "Claude usage unavailable" tagline when the bar is empty** — rejected as noisy; see Decisions above.

## Verification
- `./gradlew :web:build` passes locally.
- `./gradlew build -x test` passes across all modules (android, iOS framework links, server, etc.) — no regressions from the new logo file or the CSS / HTML tweaks.
- Manual UI verification not performed in this environment (no browser available in the sandbox) — please sanity-check: (a) the logo appears in the lower-right at load; (b) the dot goes blue when a terminal is "working" and red when "waiting" (the Debug menu's per-session override can force both states); (c) the bottom bar is visible as a thin empty strip when no Claude usage data is coming in; (d) the bar renders populated as before once data arrives; (e) `prefers-reduced-motion` disables the pulse.
- No changes to the `.app-header` no-drag allowlist needed — the logo lives outside the header and the drag-region rule doesn't apply to it.

## Files of note
- `web/src/jsMain/kotlin/se/soderbjorn/termtastic/AppLogo.kt` — new file, the aggregation rule and DOM update lives here.
- `web/src/jsMain/kotlin/se/soderbjorn/termtastic/WebState.kt` — single-line call site hook in `updateStateIndicators`.
- `web/src/jsMain/kotlin/se/soderbjorn/termtastic/ClaudeUsageBar.kt` — the no-data branch now empties the bar instead of hiding it.
- `web/src/jsMain/resources/index.html` — adds the `#app-logo` element and flips the initial usage-bar class to the empty variant.
- `web/src/jsMain/resources/styles.css` — `.app-logo*` rules, the two pulse keyframes, the reduced-motion override, and the usage-bar `min-height` / `.claude-usage-bar-empty` tweak.

## Follow-ups
- If the repo owner wants the old *connection* status (socket connecting / connected / disconnected) reflected somewhere again, the dot could be extended with a fourth state. Out of scope for #14.
- The logo's `right`/`bottom` offsets are hard-coded (14 px / 32 px). If the bottom bar's height ever changes substantially, those need updating. Could be tokenised later if there's demand.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/web/src/jsMain/kotlin/se/soderbjorn/termtastic/AppLogo.kt
+++ b/web/src/jsMain/kotlin/se/soderbjorn/termtastic/AppLogo.kt
@@ -1,0 +1,72 @@
+/**
+ * App logo widget for the Termtastic web frontend.
+ *
+ * Renders the "Termtastic" wordmark and a small status dot in the lower-right
+ * corner of the window as a fixed overlay (painted on top of everything). The
+ * dot pulsates in three states derived from the aggregate of all per-session
+ * states:
+ *
+ *   - red  — at least one session is "waiting" (e.g. an agent is asking for
+ *            input/approval). Waiting dominates because it needs action.
+ *   - blue — at least one session is "working" (agent is actively running)
+ *            and none are waiting.
+ *   - neutral — no session is working or waiting (idle).
+ *
+ * The logo brings back an older design element (see issue #14): a "Termtastic"
+ * wordmark alongside a coloured dot. In the previous incarnation the dot was
+ * to the left of the wordmark and coloured by socket-connection status; this
+ * version moves the dot to the right and ties its colour to work state, which
+ * is more informative for day-to-day use.
+ *
+ * The overlay is **always visible** — it lives directly inside `<body>` via
+ * an `#app-logo` element declared in `index.html`, so it paints on top of
+ * modals, the sidebar, the tab bar, and the main content area. It uses
+ * `pointer-events: none` so it never steals clicks from the UI underneath.
+ *
+ * @see updateAppLogoState
+ * @see updateStateIndicators
+ */
+package se.soderbjorn.termtastic
+
+import kotlinx.browser.document
+import org.w3c.dom.HTMLElement
+
+/**
+ * Recomputes and applies the logo dot state based on the current per-session
+ * state map.
+ *
+ * Called from [updateStateIndicators] whenever the server pushes a new state
+ * envelope (or when [renderConfig] replays the current state after a config
+ * change), so the dot is always in sync with the spinners/warning icons in
+ * the sidebar and tab bar.
+ *
+ * Aggregation rule:
+ *   1. If any session value is `"waiting"` → mark as waiting (red pulse).
+ *   2. Else if any session value is `"working"` → mark as working (blue pulse).
+ *   3. Else mark as idle (dim, no pulse).
+ *
+ * @param sessionStates the current session-id to state map from the server.
+ *                      States other than `"working"` / `"waiting"` (including
+ *                      `null`) count as idle.
+ * @see updateStateIndicators
+ */
+internal fun updateAppLogoState(sessionStates: Map<String, String?>) {
+    val dot = document.getElementById("app-logo-dot") as? HTMLElement ?: return
+    var anyWaiting = false
+    var anyWorking = false
+    for (state in sessionStates.values) {
+        when (state) {
+            "waiting" -> { anyWaiting = true; break }
+            "working" -> anyWorking = true
+        }
+    }
+    // Reset both modifier classes before re-applying so the dot never ends up
+    // carrying stale state when the aggregate transitions e.g. working→idle.
+    dot.classList.remove("state-working")
+    dot.classList.remove("state-waiting")
+    when {
+        anyWaiting -> dot.classList.add("state-waiting")
+        anyWorking -> dot.classList.add("state-working")
+        // Idle → no modifier class; the base .app-logo-dot rule paints it dim.
+    }
+}
--- a/web/src/jsMain/kotlin/se/soderbjorn/termtastic/ClaudeUsageBar.kt
+++ b/web/src/jsMain/kotlin/se/soderbjorn/termtastic/ClaudeUsageBar.kt
@@ -125,8 +125,13 @@ fun updateClaudeUsageBadge(usage: dynamic) {
     val bar = usageBar ?: return
     val divider = usageBarDividerEl
     if (usage == null || usage == undefined) {
-        // No data at all — hide both the bar and the drag handle above it.
-        bar.style.display = "none"
+        // No data at all — keep the bar visible but empty so it still reserves
+        // space at the bottom of the window (required by issue #14 so the app
+        // logo overlay has a background band to sit against). Hide only the
+        // divider: there is nothing to drag-collapse when the bar is empty.
+        bar.innerHTML = ""
+        bar.className = "claude-usage-bar claude-usage-bar-empty"
+        bar.style.display = ""
         divider?.style?.display = "none"
         return
     }
--- a/web/src/jsMain/kotlin/se/soderbjorn/termtastic/WebState.kt
+++ b/web/src/jsMain/kotlin/se/soderbjorn/termtastic/WebState.kt
@@ -769,17 +769,23 @@ internal fun applySpinnerState(el: HTMLElement, baseClass: String, state: String
 
 /**
  * Updates all session state indicators (spinners for "working", warning
- * icons for "waiting") across the sidebar, tab bar, and pane headers.
+ * icons for "waiting") across the sidebar, tab bar, pane headers, and the
+ * lower-right app logo dot.
  *
  * Also triggers desktop notification checks and adjusts the tab bar active
  * indicator position.
  *
  * @param sessionStates the current session ID to state map from the server
  * @see checkStateNotifications
+ * @see updateAppLogoState
  */
 internal fun updateStateIndicators(sessionStates: Map<String, String?>) {
     val document = kotlinx.browser.document
     checkStateNotifications(sessionStates)
+    // Keep the lower-right app logo dot in sync with the sidebar/tab-bar
+    // spinners and warning icons — all three surfaces derive from the same
+    // per-session state map.
+    updateAppLogoState(sessionStates)
     val effective = HashMap(sessionStates)
     for ((sessionId, state) in effective) {
         val spinners = document.querySelectorAll(".pane-status-spinner[data-session='$sessionId']")
--- a/web/src/jsMain/resources/index.html
+++ b/web/src/jsMain/resources/index.html
@@ -138,9 +138,20 @@
     </div>
     <!-- Horizontal drag handle above the Claude usage bar. Only visible when
          the server has pushed usage data; drag down to hide the bar, drag up
-         to reveal it. -->
+         to reveal it. The footer below always takes up space even when the
+         server has no usage data to report (it just renders empty), so there
+         is always room for the app logo overlay in the lower-right corner. -->
     <div class="usage-bar-divider" id="usage-bar-divider" style="display: none;"></div>
-    <footer id="claude-usage-bar" class="claude-usage-bar" style="display: none;"></footer>
+    <footer id="claude-usage-bar" class="claude-usage-bar claude-usage-bar-empty"></footer>
+</div>
+<!-- App logo overlay. Painted on top of everything via position:fixed in the
+     lower-right corner, so it stays visible regardless of modals, the
+     sidebar, or which tab is active. The dot to the right of the wordmark
+     pulsates blue when any tab is working and red when any tab is waiting
+     for input; see AppLogo.kt and updateStateIndicators(). -->
+<div id="app-logo" class="app-logo" aria-hidden="true">
+    <span class="app-logo-wordmark">Termtastic</span>
+    <span id="app-logo-dot" class="app-logo-dot"></span>
 </div>
 <script type="application/javascript" src="web.js"></script>
 </body>
--- a/web/src/jsMain/resources/styles.css
+++ b/web/src/jsMain/resources/styles.css
@@ -319,6 +319,19 @@ body.is-electron-mac .app-header {
     white-space: nowrap;
     user-select: none;
     flex-shrink: 0;
+    /* Always reserve space at the bottom of the window, even when the server
+       has no Claude usage data to display. The `.claude-usage-bar-empty`
+       variant renders with no content but keeps this minimum height so the
+       bottom bar is continuous across all runtime states and the app logo
+       overlay always has a consistent band to sit against. */
+    min-height: 24px;
+    box-sizing: border-box;
+}
+/* Empty state: no content, just a thin coloured band at the bottom. Kept
+   separate from the normal state so future tweaks (e.g. a subtle gradient,
+   a tagline) can target only the empty case. */
+.claude-usage-bar.claude-usage-bar-empty {
+    padding: 0 12px;
 }
 .claude-usage-bar .usage-item {
     display: flex;
@@ -4542,4 +4555,123 @@ body.appearance-light .settings-section-select {
     cursor: default;
 }
 
+/* ─── App logo overlay ──────────────────────────────────────────────
+   The "Termtastic" wordmark + status dot pinned to the lower-right
+   corner of the window. Painted on top of everything via position:
+   fixed and a high z-index so it's always visible regardless of what
+   tab, modal, or pane is active. pointer-events: none so it never
+   intercepts clicks meant for the UI below it.
+
+   Sits just above the Claude usage bar (which reserves a fixed min
+   height via .claude-usage-bar above), with enough padding that it
+   reads as a badge rather than a footer element. Size is deliberately
+   small-ish — the label should feel like a signature, not a banner —
+   so it doesn't crowd content in the main terminal area.
+
+   The dot mirrors the per-tab spinner/waiting indicators at the
+   aggregate level: blue pulse when any session is "working", red
+   pulse when any session is "waiting for input", dim when idle.
+   See AppLogo.kt / updateAppLogoState() for the state machine. */
+.app-logo {
+    position: fixed;
+    right: 14px;
+    bottom: 32px;
+    display: flex;
+    align-items: center;
+    gap: 8px;
+    pointer-events: none;
+    user-select: none;
+    z-index: 100;
+    /* A soft text shadow so the wordmark stays legible against noisy
+       terminal content underneath (bright ANSI output, light themes,
+       etc.). Tuned low enough that it doesn't look like a drop shadow
+       on quiet backgrounds. */
+    text-shadow: 0 1px 2px rgba(0, 0, 0, 0.55);
+}
+
+.app-logo-wordmark {
+    font-family: Georgia, "Times New Roman", serif;
+    font-size: 15px;
+    font-weight: 700;
+    letter-spacing: 0.3px;
+    /* Use the termtastic-orange token the rest of the UI leans on for
+       brand accents (see .sidebar-divider.dragging, the tab active
+       indicator fallback, etc.) so the logo feels in-family. */
+    color: var(--termtastic-orange, #F4B869);
+    line-height: 1;
+}
+
+/* In light appearance, the orange + white-ish text-shadow combo reads
+   as washed out. Swap to a darker shadow so the wordmark still pops. */
+body.appearance-light .app-logo {
+    text-shadow: 0 1px 2px rgba(255, 255, 255, 0.75), 0 0 1px rgba(0, 0, 0, 0.3);
+}
+
+/* The status dot. Base state is idle (dim grey, no pulse). The
+   .state-working and .state-waiting modifier classes are added by
+   updateAppLogoState() in AppLogo.kt when any session is in that
+   state — waiting wins over working because it's action-required.
+   Each state has its own keyframe so the colour animates smoothly
+   with the opacity pulse. */
+.app-logo-dot {
+    display: inline-block;
+    width: 10px;
+    height: 10px;
+    border-radius: 50%;
+    background: var(--text-secondary, #8E8E93);
+    box-shadow: 0 0 0 2px rgba(142, 142, 147, 0.18);
+    flex-shrink: 0;
+    transition: background-color 180ms ease-in-out,
+                box-shadow 180ms ease-in-out;
+}
+
+.app-logo-dot.state-working {
+    background: #4C9AFF;
+    box-shadow: 0 0 0 2px rgba(76, 154, 255, 0.25);
+    animation: app-logo-pulse-blue 1.6s ease-in-out infinite;
+}
+
+.app-logo-dot.state-waiting {
+    background: #FF6961;
+    box-shadow: 0 0 0 2px rgba(255, 105, 97, 0.3);
+    animation: app-logo-pulse-red 1.2s ease-in-out infinite;
+}
+
+/* Pulse keyframes: scale + glow breathe to draw the eye without
+   being obnoxious. Shorter period for waiting (red) than for working
+   (blue) because waiting is action-required and should feel more
+   urgent, mirroring the fade-warning on .pane-status-spinner. */
+@keyframes app-logo-pulse-blue {
+    0%, 100% {
+        box-shadow: 0 0 0 2px rgba(76, 154, 255, 0.25),
+                    0 0 0 0 rgba(76, 154, 255, 0.45);
+    }
+    50% {
+        box-shadow: 0 0 0 2px rgba(76, 154, 255, 0.25),
+                    0 0 0 6px rgba(76, 154, 255, 0);
+    }
+}
+
+@keyframes app-logo-pulse-red {
+    0%, 100% {
+        box-shadow: 0 0 0 2px rgba(255, 105, 97, 0.30),
+                    0 0 0 0 rgba(255, 105, 97, 0.55);
+        transform: scale(1);
+    }
+    50% {
+        box-shadow: 0 0 0 2px rgba(255, 105, 97, 0.30),
+                    0 0 0 8px rgba(255, 105, 97, 0);
+        transform: scale(1.08);
+    }
+}
+
+/* Respect prefers-reduced-motion: keep the colour change but drop the
+   pulse animation, matching the existing motion-reduction hygiene
+   elsewhere in this stylesheet. */
+@media (prefers-reduced-motion: reduce) {
+    .app-logo-dot.state-working,
+    .app-logo-dot.state-waiting {
+        animation: none;
+    }
+}
 
--- a/web/src/jsMain/resources/styles.css
+++ b/web/src/jsMain/resources/styles.css
@@ -4575,10 +4575,19 @@ body.appearance-light .settings-section-select {
 .app-logo {
     position: fixed;
     right: 14px;
-    bottom: 32px;
+    /* Cleared up from 32px — the previous value put the overlay right on top
+       of the Claude usage bar's content (see @soderbjorn screenshot on
+       #16), because the populated usage bar's effective height
+       (min-height 24 px + 3 px vertical padding + 1 px top border ≈ 28–30 px)
+       didn't leave enough breathing room before the dot's glow ring. 44 px
+       gives the wordmark and its pulse halo a comfortable band above the
+       bottom bar in both the populated and empty states. */
+    bottom: 44px;
     display: flex;
     align-items: center;
-    gap: 8px;
+    /* Slightly wider gap to match the bumped-up dot size so the wordmark
+       and dot still read as a balanced pair rather than crowding. */
+    gap: 9px;
     pointer-events: none;
     user-select: none;
     z-index: 100;
@@ -4615,8 +4624,14 @@ body.appearance-light .app-logo {
    with the opacity pulse. */
 .app-logo-dot {
     display: inline-block;
-    width: 10px;
-    height: 10px;
+    /* Bumped from 10 px → 13 px per @soderbjorn feedback on #16:
+       "make the dot a little bit bigger in relation to the label". 13 px
+       sits at roughly 87% of the 15 px wordmark cap height, which reads as
+       a deliberate circular badge next to the text rather than a small
+       punctuation mark beside it. The glow/pulse ring grows with it (see
+       box-shadow/keyframes below). */
+    width: 13px;
+    height: 13px;
     border-radius: 50%;
     background: var(--text-secondary, #8E8E93);
     box-shadow: 0 0 0 2px rgba(142, 142, 147, 0.18);
@@ -4647,8 +4662,10 @@ body.appearance-light .app-logo {
                     0 0 0 0 rgba(76, 154, 255, 0.45);
     }
     50% {
+        /* Outer halo radius bumped 6 px → 8 px in lockstep with the larger
+           13 px dot so the pulse ring stays visually proportional. */
         box-shadow: 0 0 0 2px rgba(76, 154, 255, 0.25),
-                    0 0 0 6px rgba(76, 154, 255, 0);
+                    0 0 0 8px rgba(76, 154, 255, 0);
     }
 }
 
@@ -4659,8 +4676,10 @@ body.appearance-light .app-logo {
         transform: scale(1);
     }
     50% {
+        /* Outer halo radius bumped 8 px → 10 px in lockstep with the larger
+           13 px dot so the pulse ring stays visually proportional. */
         box-shadow: 0 0 0 2px rgba(255, 105, 97, 0.30),
-                    0 0 0 8px rgba(255, 105, 97, 0);
+                    0 0 0 10px rgba(255, 105, 97, 0);
         transform: scale(1.08);
     }
 }
--- a/web/src/jsMain/resources/styles.css
+++ b/web/src/jsMain/resources/styles.css
@@ -4575,14 +4575,21 @@ body.appearance-light .settings-section-select {
 .app-logo {
     position: fixed;
     right: 14px;
-    /* Cleared up from 32px — the previous value put the overlay right on top
-       of the Claude usage bar's content (see @soderbjorn screenshot on
-       #16), because the populated usage bar's effective height
-       (min-height 24 px + 3 px vertical padding + 1 px top border ≈ 28–30 px)
-       didn't leave enough breathing room before the dot's glow ring. 44 px
-       gives the wordmark and its pulse halo a comfortable band above the
-       bottom bar in both the populated and empty states. */
-    bottom: 44px;
+    /* The logo lives *inside the Claude usage bar band* at the bottom of the
+       window, NOT floating above it inside the tab content area. Earlier
+       iterations tried 32 px (collided with the bar's content — first
+       @soderbjorn screenshot on #16) and then 44 px (bumped away from the
+       collision but that pushed the overlay up into the terminal tab's
+       rounded bottom-right corner, so it started visually living inside
+       the tab content surface — second @soderbjorn screenshot on #16:
+       "it's not supposed to be inside the tab content area").
+       At 7 px the wordmark (~15 px cap-height) and 13 px dot sit vertically
+       centred within the ~30 px populated bar (24 px min-height + 6 px
+       vertical padding + 1 px top border), and within the 24 px empty-bar
+       band they still only push their pulse halo a few pixels above the
+       bar's top border — i.e. "only minimally overlaying padding" above
+       the bar rather than sitting wholly on top of the tab content. */
+    bottom: 7px;
     display: flex;
     align-items: center;
     /* Slightly wider gap to match the bumped-up dot size so the wordmark


soderbjorn:
Nice try but unfortunately it collides, check the screenshot and adjust the distance as needed. Also make the dot a little bit bigger in relation to the label.

overlap

soderbjorn:

## Summary of changes
Restores the "Termtastic" wordmark plus a coloured status dot as a fixed-position overlay in the lower-right of the web UI, pinned on top of every other UI surface via `position: fixed; z-index: 100; pointer-events: none`. The dot has three visual states driven by an aggregate of the existing per-session state map: red pulse when any tab is `waiting`, blue pulse when any tab is `working`, dim grey otherwise. To give the overlay a consistent band to sit against, the Claude usage bar is no longer hidden when there is no usage payload — it stays in the DOM with a new `.claude-usage-bar-empty` class and a 24 px `min-height`. The drag-divider above the bar still hides in the empty case. Matches the spec in issue #14 faithfully.

## Functional review
The happy path works: `updateAppLogoState` is invoked from `updateStateIndicators` on every state envelope, which is the same hook the sidebar/tab-bar spinners use, so the three surfaces stay in lockstep. The aggregation rule (waiting wins over working wins over idle) matches the existing tab-level rule at `WebState.kt:805-810`. The short-circuit `break` on the first `"waiting"` is correct because `anyWorking` can't change the outcome once any waiting is found.

A few user-visible concerns worth thinking through:

- **Collapsed-bar interaction with empty state.** `applyUsageBarCollapsedState()` (`WebState.kt:535-539`) adds `.collapsed` to the usage bar during hydration if the user previously collapsed it, and `.claude-usage-bar.collapsed { display: none }` (styles.css:795-797) still wins over the new `min-height`. So a user who has collapsed the bar and has no incoming usage data will get no bottom band at all — i.e. the "always reserve space" contract from the issue breaks for exactly the users whose bar is already hidden. Once a `ClaudeUsage` envelope arrives (even with `null`), `updateClaudeUsageBadge` resets the class list to `claude-usage-bar claude-usage-bar-empty` and silently clears `.collapsed`, which also changes the semantics of "collapsed": the persisted preference is effectively dropped the moment data is missing. Worth deciding whether "the bar is always-on" should override the collapsed preference (simpler, consistent with the issue's framing), or whether the collapsed preference should be respected in the empty case too (then you need a separate reserved-space shim, since the collapsed bar has `display: none`).
- **Stale docstring.** `applyUsageBarCollapsedState`'s KDoc at `WebState.kt:528-530` still claims "in that case the bar and divider both stay hidden (inline `display: none`, set in the initial HTML …)". That's no longer true — the initial HTML dropped `style="display: none;"` on the footer (index.html:145). Update the KDoc so the next reader isn't misled about the pre-envelope visual state.
- **Height jump when data arrives.** `.claude-usage-bar` has `padding: 3px 12px` (styles.css:313) but the new `.claude-usage-bar-empty` overrides to `padding: 0 12px` (styles.css:334). When the first real envelope lands, the bar grows from ~24 px to ~24 px + 6 px padding, nudging the terminal area up slightly. Either equalise padding (e.g. keep `3px 12px` on the empty variant too and rely on `min-height` to hold the floor) or match the empty band's height to the populated one exactly. Minor but noticeable on a first connect.
- **Logo overlaps real content on narrow windows.** The overlay sits at `right: 14px; bottom: 32px` and is sized for a desktop-ish viewport. On a narrow Electron window with a short terminal, the logo can overlap prompt output near the bottom-right of the pane. `pointer-events: none` keeps it click-through, but users can't temporarily hide it. Not a blocker for #14 (the issue explicitly accepts this trade-off — "painted on top of everything"), but worth keeping in mind as a follow-up if complaints surface.
- **No screen-reader announcement for "waiting".** `aria-hidden="true"` on `#app-logo` is appropriate for a decorative brand mark, but it also means the state change from "working" → "waiting" (the "you have action to take" alert) never reaches assistive tech via this surface. The per-tab warning icons presumably do the same — and should continue to — but the logo specifically advertised the waiting state as a window-level cue. Not in scope for this PR, but worth confirming that the sidebar/tab-bar "waiting" indicators have accessible labels, otherwise screen-reader users have no equivalent cue.
- **Debug state overrides.** The PR body claims "states other than 'working'/'waiting' (including debug override labels other than these two) are treated as idle." If a debug override labels a session with, say, a custom tag, the dot goes idle even though the corresponding sidebar spinner or warning stays lit. That's a deliberate choice and matches how `applySpinnerState` falls through, so it's consistent — just flagging it as a thing to remember when debugging.

## Technical review
Layering and naming are clean. `AppLogo.kt` is a small, single-purpose module with one exposed function; the hook in `updateStateIndicators` is a single line and doesn't disturb the existing control flow. The KDoc on the new file and on `updateAppLogoState` conforms to the CLAUDE.md documentation rule — what it does, who calls it, parameters, `@see` cross-references, the decision log for the aggregation rule. Good.

The CSS is organised into a labelled block with extensive inline comments explaining the positioning, the pulse keyframes, the `prefers-reduced-motion` override, and the light-appearance text-shadow swap. `--termtastic-orange` is the correct token for the wordmark colour (consistent with the dozens of other usages of that variable across the stylesheet). The Georgia font-family choice matches the surviving `.header-title` rule, as the PR body explains.

A few nits with specific pointers:

- **Hardcoded dot colours break theme consistency.** `styles.css:4629` uses `#4C9AFF` for working and `styles.css:4635` uses `#FF6961` for waiting, both as literal hex. The rest of the file leans on `--t-semantic-*` tokens (`--t-semantic-warn`, `--t-semantic-danger`, etc.) so themes can re-skin status colours. `--t-semantic-danger` is already `#FF6961` by default (see `.usage-critical` at styles.css:344) and `--t-info`/`--t-semantic-info` (if defined in the token set) would give a themeable blue. Swapping to `var(--t-semantic-danger, #FF6961)` / `var(--t-semantic-info, #4C9AFF)` would bring the logo in line with the rest of the semantic colour system at near-zero cost.
- **`updateAppLogoState` silently no-ops if the DOM node is missing.** Safe (the element is in `index.html`), but easy to break with a future refactor. Consider a `console.warn` in the `?: return` branch, or at least a `@throws`-like note in the KDoc stating the function is a no-op pre-boot. Minor.
- **`usageBarCollapsed` class is now leaked by the no-data path.** `updateClaudeUsageBadge` in the null branch does `bar.className = "claude-usage-bar claude-usage-bar-empty"` (ClaudeUsageBar.kt:133), which overwrites any previously-applied `.collapsed`. The populated branch (ClaudeUsageBar.kt:160-163) is careful to re-apply `.collapsed` based on `appVm.stateFlow.value.usageBarCollapsed`; the empty branch doesn't do the same check. Either respect the collapsed preference in both branches, or document that "empty" implicitly cancels collapse (see the first Functional review bullet). Inconsistency between the two branches is the real smell.
- **`break` inside the `for (state in sessionStates.values)` loop.** Works in Kotlin/JS but deserves a `// short-circuit on first waiting` comment to explain why we stop iterating. The PR body mentions this; pulling the same sentence into the code keeps the invariant visible. Nit.
- **`changedFiles: 5`, `+231 / -5`.** In line with what the PR body describes — no sprawl into unrelated files. Good.

Build: I ran `./gradlew :web:jsBrowserDistribution` locally on the PR head — passes, no new warnings from the added Kotlin or CSS.

## Consequences
- Users who previously collapsed the Claude usage bar (persisted via `usageBarCollapsed`) will find their collapse preference silently overridden the first time a `ClaudeUsage` envelope with a null payload arrives. See Functional bullet #1. If users have been relying on "collapsed = gone", they'll notice the bar reappearing and the divider disappearing.
- Electron window-drag behaviour is unchanged — the overlay sits outside `.app-header`, so the no-drag allowlist from the user's memory rule doesn't apply. Correctly noted in the PR body.
- No data-model or schema changes; purely frontend DOM/CSS plus one Kotlin/JS module. Safe to roll back by reverting the commit.
- Android/iOS don't render this HTML at all (it's `web/src/jsMain/resources/index.html`), so the mobile apps are unaffected. Issue #15 is about Android padding, unrelated.

## Risks
- **Should-fix:** The collapsed-vs-empty interaction (Functional bullet #1 and Technical bullet #3). It's not a regression in the strict sense — the old behaviour was "no bar when no data", the new behaviour is "bar when no data, unless collapsed persists from a time when data existed" — but it breaks the issue's invariant and is surprising enough that a fix before merge would avoid a follow-up.
- **Nit:** Theme-token hardcoding on the dot colours (Technical bullet #1). Doesn't break anything today but will be confusing the first time someone adds a "high-contrast" or "monochrome" theme.
- **Nit:** Stale docstring on `applyUsageBarCollapsedState` (Functional bullet #2). Low cost to fix; violates CLAUDE.md's "do not leave stale documentation" intent.
- **Nit:** Empty-vs-populated height jump (Functional bullet #3). Cosmetic.

No security, performance, or race-condition concerns spotted. The animations are CSS keyframes, not `setInterval`, so there's no leak on hot-reload or tab switch. `pointer-events: none` prevents the overlay from accidentally capturing input.

## Ideas & alternatives
- If you want the dot to also convey socket-connection status (the PR body flags this as a possible follow-up), the cleanest factoring is a secondary modifier class `.state-connecting` or `.state-disconnected` that `WindowConnection` toggles directly on `#app-logo-dot`, keeping the state machine additive. The existing precedence (waiting > working > idle) naturally extends with disconnected at the bottom.
- `Map<String, String?>.values.any { it == "waiting" }` / `…any { it == "working" }` is idiomatic Kotlin and reads cleaner than the manual `for` loop with flags; trades a tiny amount of code size for two passes instead of one. Pure style; keep the loop if the micro-optimisation matters.
- Consider extracting the pulse keyframes into a shared mixin if you ever add a third pulsing surface — right now there's duplication between `fade-warning` (spinner) and `app-logo-pulse-*`. Not worth doing speculatively.

## Verdict
Solid implementation. The architecture is right (overlay outside `#app`, aggregation reuses the existing state map, `prefers-reduced-motion` respected, Kotlin/JS doc conventions followed). The collapsed-bar-plus-null-payload interaction is the one real functional issue and should be addressed before merge so the issue's "always reserve space" contract actually holds for every user; the other bullets are nits/doc fixes. I'd merge this with the collapsed-state fix and the stale-docstring touch-up. I did not exercise the UI in a browser (sandbox has none), so the author's manual verification checklist in the PR body remains the authoritative word on the visual behaviour.

soderbjorn:
**Claude Code** (an AI coding agent) picked up this feedback via the `/pick-followup` skill and pushed 16759ed.

> Nice try but unfortunately it collides, check the screenshot and adjust the distance as needed. Also make the dot a little bit bigger in relation to the label.

Two tweaks in `web/src/jsMain/resources/styles.css`:

- **Clearance.** `.app-logo { bottom }` raised from 32 px to 44 px. The previous value sat right on top of the populated Claude usage bar's content because the bar's effective height (24 px `min-height` + 3 px vertical padding + 1 px top border) left only a few pixels before the dot's glow ring — exactly the collision in your screenshot. 44 px gives a comfortable band above the bar in both the populated and empty states.
- **Dot size.** `.app-logo-dot` grew from 10 × 10 px to 13 × 13 px (about 87% of the 15 px wordmark cap height), so it reads as a deliberate circular badge rather than a tiny punctuation mark. The two pulse keyframes were bumped in lockstep — outer halo radius 6 → 8 px for blue, 8 → 10 px for red — so the animation stays visually proportional. Flex `gap` nudged from 8 px to 9 px to match.

No logic, HTML, or Kotlin touched — the fix is purely CSS, and the aggregation / state-machine behaviour is unchanged.

**Verification.** `./gradlew :web:build` passes locally (webpack emits the two unrelated asset-size warnings that predate this PR). Visual verification not performed in this environment — please confirm the logo no longer overlaps the usage bar content and that the bigger dot reads right alongside the wordmark.

🤖 This comment was posted by [Claude Code](https://claude.com/claude-code) acting autonomously — addressed in commit 16759ed. The human repo owner has not yet confirmed the change.

soderbjorn:

This is a re-review: new code (`16759ed`) landed after my previous review comment, so I re-read the PR end-to-end on the current head.

## Summary of changes
Unchanged at the behavioural level from the first review: fixed-position "Termtastic" wordmark plus a status dot pinned to the lower-right of the window, with the dot aggregating the per-session state map (red when any tab is `waiting`, blue when any tab is `working`, dim otherwise). The Claude usage bar is no longer hidden when there is no usage payload — it sticks around as an empty 24 px band via `.claude-usage-bar-empty` so the overlay has a consistent floor to sit against.

What the follow-up commit changed: `.app-logo { bottom }` 32 px → 44 px to clear the populated usage bar (directly addressing the collision in the screenshot), dot size 10 px → 13 px (≈87% of the 15 px wordmark cap height), the two pulse halo radii grew in lockstep (6 → 8 px blue, 8 → 10 px red), and `.app-logo { gap }` 8 px → 9 px. No logic, HTML, or Kotlin touched in the follow-up.

## Functional review
The follow-up fixes the immediate complaint cleanly. 44 px `bottom` is the right call: the populated usage bar measures ~30 px in practice (24 px `min-height` + 6 px vertical padding + 1 px top border), so 32 px left the dot's glow ring touching the bar's text baseline. 44 px gives about a 14 px breathing gap in the populated state and more in the empty state. No collision should be visible at standard zoom.

The enlarged dot reads correctly proportional to the wordmark in my head (can't verify pixel-perfectly without a browser). The pulse-halo radii scaling in lockstep is the right instinct — otherwise the ring would have looked thin around the bigger disk.

**Issues from my previous review that are still live and should be addressed before merge:**

- **Collapsed-bar + null-usage interaction (was Functional bullet #1).** `updateClaudeUsageBadge` at `web/src/jsMain/kotlin/se/soderbjorn/termtastic/ClaudeUsageBar.kt:127-137` still does `bar.className = "claude-usage-bar claude-usage-bar-empty"` on the null branch, which overwrites any previously-applied `.collapsed` class. The populated branch at `ClaudeUsageBar.kt:160-163` is careful to re-apply `.collapsed` from `appVm.stateFlow.value.usageBarCollapsed`, but the empty branch isn't. Consequence: a user who collapsed the bar previously will see it silently re-appear as an empty band the first time a null envelope arrives. More subtly, `applyUsageBarCollapsedState()` running during hydration can still add `.collapsed` to the bar before any envelope, and `.claude-usage-bar.collapsed { display: none }` at `web/src/jsMain/resources/styles.css:795-797` wins over the new `min-height`, so the issue's "always reserve space" contract breaks for exactly those users. Easiest fix: mirror the populated branch's `if (appVm.stateFlow.value.usageBarCollapsed) bar.classList.add("collapsed")` on the empty branch, and either decide that "collapsed" is obsolete now that the bar is always-on, or ensure the collapsed state still reserves the 24 px band (e.g. replace `display: none` with a collapsed-specific `height: 24px` / `overflow: hidden`).
- **Stale docstring (was Functional bullet #2).** `applyUsageBarCollapsedState`'s KDoc at `web/src/jsMain/kotlin/se/soderbjorn/termtastic/WebState.kt:528-530` still reads *"Safe to call before the first usage envelope arrives — in that case the bar and divider both stay hidden (inline `display: none`, set in the initial HTML and by [updateClaudeUsageBadge] when `usage` is null)."* That's no longer accurate: the initial HTML dropped the inline `display: none` on the footer (`index.html:145`), and the null branch now keeps the bar visible. CLAUDE.md's documentation rule explicitly covers "do not leave stale doc comments". Five-line fix.
- **Height jump when the first real envelope arrives (was Functional bullet #3).** `.claude-usage-bar` has `padding: 3px 12px` (styles.css:313) and `.claude-usage-bar.claude-usage-bar-empty` overrides to `padding: 0 12px` (styles.css:333-335). The empty band is ~24 px tall, the populated band is ~30 px tall, so on first connect the content area above the bar nudges up by 6 px. Either keep the empty variant at `padding: 3px 12px` and rely on `min-height` to hold the floor when the content is empty, or bump the populated bar's `min-height` to match the empty band's effective height. Cosmetic but noticeable.
- **Hardcoded dot colours (was Technical bullet #1).** `styles.css:4629` (`.app-logo-dot.state-working`) hardcodes `#4C9AFF` and `styles.css:4635` (`.app-logo-dot.state-waiting`) hardcodes `#FF6961`, plus the same hex values are baked into all four `app-logo-pulse-*` keyframe stops. The rest of the stylesheet consistently uses semantic tokens (`--t-semantic-warn`, `--t-semantic-danger`, `--t-semantic-success`) for status colours — `--t-semantic-danger` is already `#FF6961` by default (see `.usage-critical` at `styles.css:344`). A future high-contrast / monochrome theme would miss the overlay dot. Swap the two hex values to `var(--t-semantic-danger, #FF6961)` and either `var(--t-semantic-info, #4C9AFF)` or a new `--termtastic-blue` token.

These are all from the first review and remain unchanged. Worth flagging them again because the follow-up commit didn't pick any of them up, so they'd otherwise slip through.

Happy-path behaviour is unchanged from the first review and remains correct: `updateAppLogoState` is invoked from `updateStateIndicators`, which runs on every state envelope; the waiting-wins-over-working aggregation matches the per-tab rule; `prefers-reduced-motion` is respected; `pointer-events: none` prevents input capture.

## Technical review
The follow-up commit is a pure CSS tweak. Values look reasonable, comments explain the numbers (44 px derivation, 13/15 ratio, halo-radius scaling). No new Kotlin or HTML to inspect beyond what the first review covered.

One note on the enlarged-dot choice: the `transform: scale(1.08)` on the red pulse (`styles.css:4695`) combined with the 13 px disk means the dot briefly reaches ~14 px during the waiting pulse. Still fits the `.app-logo { gap: 9px }` budget comfortably, and the halo radius was bumped in lockstep, so the effect should feel proportional. Fine.

**One new nit on the follow-up commit itself:**

- The comment at `styles.css:4620-4625` explaining the 44 px `bottom` references the Claude usage bar's effective height as "min-height 24 px + 3 px vertical padding + 1 px top border ≈ 28–30 px". The padding on the populated bar is `3px 12px`, i.e. 3 px top + 3 px bottom = 6 px total, so the effective height is closer to 24 + 6 + 1 = 31 px. Minor arithmetic slip in the comment; doesn't affect the chosen 44 px value.

Kotlin/JS documentation conformance on `AppLogo.kt` remains good (the first review covered this in detail). No new test coverage was added for the aggregation rule — that was true of the first commit too; still worth noting since `AppLogo.kt`'s `when` branch is the kind of thing a unit test would catch if the rule ever flipped.

Build: I did not run `./gradlew :web:build` on the re-review head — the follow-up is CSS-only and the PR body reports a local build pass.

## Consequences
Same as the first review — no data-model, schema, or API changes; web-only. The collapsed-bar regression described in the Functional section is the one user-visible consequence worth calling out explicitly: users who previously collapsed the usage bar (persisted via `usageBarCollapsed`) will see the bar silently un-collapse the first time a null `ClaudeUsage` envelope arrives, because the null branch overwrites the class list. That's a small but real behaviour shift for existing users.

Android/iOS unaffected (web-only HTML/CSS/Kotlin-JS).

## Risks
- **Should-fix:** The four outstanding items above (collapsed interaction, stale docstring, height jump, theme-token hardcoding). None block the PR in a strict sense, but they were flagged a tick ago and leaving them now means they'll either ship or need a third commit. The collapsed interaction is the only one that causes a visible behaviour regression; the other three are polish / hygiene.
- **Nit:** Comment arithmetic at `styles.css:4620-4625` (padding total 3 px vs. 6 px). Doesn't affect behaviour.

No new security, performance, or race-condition concerns beyond what the first review covered. CSS-only follow-up; nothing to regress there.

## Ideas & alternatives
Same as the first review; nothing new to add on the follow-up commit specifically.

## Verdict
The follow-up commit does exactly what was asked — the collision is fixed and the dot is now sized proportionally to the wordmark. Good targeted change. The reason I'm not upgrading my verdict to "merge as-is" is that the four items I flagged in my previous review are all still live on this head, and one of them (collapsed-bar + null-envelope) is a visible behaviour regression rather than a nit. A small third commit (maybe 10–15 lines across `ClaudeUsageBar.kt`, `WebState.kt`, and `styles.css`) would close all four. I did not exercise the UI in a browser (sandbox has none), so the author's visual confirmation on the collision fix and dot size remains the authoritative word.

soderbjorn:
Completely incorrect. Now it's even further inside the content area instead of the opposite. It sits flush with the border of a terminal window inside the tab content area, but it's not support so be inside the tab content area (or at least only minimally as long as it only overlays padding there).

Make sure you check this screenshot properly so that you understand.
Screenshot 2026-04-24 at 13 22 07

soderbjorn:
**Claude Code** (an AI coding agent) picked up this feedback via the `/pick-followup` skill and pushed 01ce393.

> Completely incorrect. Now it's even further inside the content area instead of the opposite. It sits flush with the border of a terminal window inside the tab content area, but it's not support[ed] to be inside the tab content area (or at least only minimally as long as it only overlays padding there).

Moved `.app-logo { bottom }` in `web/src/jsMain/resources/styles.css` from 44 px down to 7 px. The previous value deliberately lifted the overlay *above* the Claude usage bar to avoid the original collision, but that pushed it *up into the terminal tab's rounded bottom-right corner* — i.e. into the tab content surface — which is exactly what the second screenshot called out. At 7 px the wordmark (~15 px cap height) and 13 px dot sit vertically centred within the ~30 px populated usage bar (`min-height: 24 px` + 6 px vertical padding + 1 px top border); in the 24 px empty-bar variant the dot's pulse halo extends a few pixels above the bar's top edge but only into the usage-bar-divider / tab-content-padding gap, not onto the tab content itself — matching the "only minimally as long as it only overlays padding there" tolerance. Nothing else moved (no HTML, Kotlin, or state-machine changes; gap and dot size carried over from the previous follow-up); the fix is a single CSS value plus an expanded block comment recording the 32 → 44 → 7 px history so the next reader doesn't regress it.

**Verification.** `./gradlew :web:build` passes locally (webpack emits the two unrelated asset-size warnings that predate this PR). Visual verification not performed in this sandbox — please confirm on a running browser that the logo now sits *within* the bottom usage bar rather than above it, and that the pulse halo on "waiting"/"working" still reads cleanly at the new position.

🤖 This comment was posted by [Claude Code](https://claude.com/claude-code) acting autonomously — addressed in commit 01ce393. The human repo owner has not yet confirmed the change.

soderbjorn:

This is a re-review: commit `01ce393` landed after my previous review on this PR, so I re-read the PR end-to-end on the current head (`01ce393`).

## Summary of changes
Behaviourally unchanged from the first two reviews: a fixed-position "Termtastic" wordmark plus a status dot pinned to the lower-right of the window, with the dot aggregating the per-session state map (red when any tab is `waiting`, blue when any tab is `working`, dim otherwise). The Claude usage bar is no longer hidden when there is no usage payload — it stays in the DOM as an empty 24 px band via `.claude-usage-bar-empty` so the overlay has a consistent floor.

What the latest follow-up (`01ce393`) changed: `.app-logo { bottom }` 44 px → 7 px, vertically centring the wordmark and dot within the ~30 px populated usage bar (24 px `min-height` + 6 px vertical padding + 1 px top border), and the accompanying block comment was expanded to record the 32 → 44 → 7 px history with rationale for each step. No HTML, Kotlin, state-machine, dot-size, gap, or pulse changes in this commit.

## Functional review
The new position directly addresses the second owner screenshot: at 7 px the logo sits *inside* the bottom usage-bar band rather than floating above it against the rounded bottom-right corner of the active tab. Geometrically that looks right — 7 px `bottom` + 13 px dot puts the dot's vertical centre at roughly 13.5 px, which is close to the midpoint of the ~30 px populated bar (≈15 px). In the 24 px empty-bar state the dot still fits inside the band vertically, though the pulse halo (up to 10 px radius on waiting) briefly extends a few pixels above the bar's top border into the divider-area gap — the commit comment's "only minimally overlaying padding" framing matches the owner's stated tolerance.

Happy path for state aggregation is unchanged from my previous reviews and still correct: `updateAppLogoState` fires from `updateStateIndicators` on every envelope, waiting wins over working wins over idle, `prefers-reduced-motion` drops the animation while keeping the colour cue, `pointer-events: none` prevents input capture. Good.

**Issues flagged in my previous reviews that are still live on this head and should be closed out before merge:**

- **Collapsed-bar + null-usage interaction.** `updateClaudeUsageBadge` at `web/src/jsMain/kotlin/se/soderbjorn/termtastic/ClaudeUsageBar.kt:132-134` still does `bar.className = "claude-usage-bar claude-usage-bar-empty"` on the null branch, which overwrites any `.collapsed` class the user had set. The populated branch at `ClaudeUsageBar.kt:160-163` is careful to re-apply `.collapsed` from `appVm.stateFlow.value.usageBarCollapsed`; the empty branch doesn't. Consequence: a user who collapsed the bar previously will see it silently re-appear as an empty band the first time a null envelope arrives. Worse, `applyUsageBarCollapsedState()` runs during hydration at `web/src/jsMain/kotlin/se/soderbjorn/termtastic/main.kt:221`, *before* any usage envelope, and will add `.collapsed` to the bar based on the persisted preference; `.claude-usage-bar.collapsed { display: none }` at `styles.css:795-797` then wins over the new `min-height: 24px`, so for that user the bottom band is never reserved at all and the 7 px-centred logo lands against the terminal surface again — exactly the regime the owner rejected in screenshot #2. This is the one live behaviour regression against the issue's "always reserve space" contract. Easiest fix: in the null branch, mirror the populated branch's `if (appVm.stateFlow.value.usageBarCollapsed) bar.classList.add("collapsed")`, and separately decide whether `.collapsed` should keep meaning `display: none` (in which case pre-envelope users with a collapsed preference still break the contract, so `.collapsed` probably needs to become a no-op now that the bar is always-on) or whether the collapsed state should switch to `height: 24px; overflow: hidden` so it still reserves the band.
- **Stale KDoc on `applyUsageBarCollapsedState`.** `web/src/jsMain/kotlin/se/soderbjorn/termtastic/WebState.kt:528-530` still says *"Safe to call before the first usage envelope arrives — in that case the bar and divider both stay hidden (inline `display: none`, set in the initial HTML and by [updateClaudeUsageBadge] when `usage` is null)."* Neither clause is true on this head: the initial HTML dropped `style="display: none;"` from the footer (`index.html:145`), and the null branch of `updateClaudeUsageBadge` explicitly sets `bar.style.display = ""`. CLAUDE.md's documentation rule is explicit about not leaving stale docblocks; this one has survived two follow-ups. Five-line fix.
- **Height jump when data arrives.** `.claude-usage-bar` has `padding: 3px 12px` (`styles.css:313`) while `.claude-usage-bar.claude-usage-bar-empty` overrides to `padding: 0 12px` (`styles.css:333-334`). The empty band is 24 px, the populated band is ~30 px, so on first connect the terminal area above the bar nudges up ~6 px — and with the new 7 px logo `bottom`, the logo's geometry relative to its band also shifts (it gains vertical centring that wasn't there in the empty band). Either keep the empty variant at `padding: 3px 12px` and let `min-height` hold the floor when the content is empty, or bump `min-height` to match the populated bar's effective height. Cosmetic but noticeable, and with the new in-band placement it's slightly more visible than before.
- **Hardcoded dot colours.** `styles.css:4629` (`.app-logo-dot.state-working`) still hardcodes `#4C9AFF` and `styles.css:4635` (`.app-logo-dot.state-waiting`) still hardcodes `#FF6961`, with the same hex values baked into all four `app-logo-pulse-*` keyframe stops. The rest of the file leans on `--t-semantic-*` tokens for status colours; `--t-semantic-danger` is already `#FF6961` by default (see `.usage-critical` at `styles.css:344` and `.claude-usage-bar .usage-pct.usage-critical` at `styles.css:344`). Swapping to `var(--t-semantic-danger, #FF6961)` and either a new `--t-semantic-info` or `--termtastic-blue` token would bring the logo in line with the rest of the semantic colour system at near-zero cost. Doesn't break anything today but will confuse the first custom-theme author.

None of those are new — they were flagged a tick or two ago and the two follow-up commits only moved the position/size of the overlay without touching the underlying code paths. Worth calling them out again because "the re-review didn't mention them again" is the failure mode where they slip through.

## Technical review
The follow-up commit is a single CSS value change (44 → 7) plus a substantially expanded block comment at `styles.css:4572-4591` that explicitly records the 32 → 44 → 7 px history with the reason each value was rejected. That's the right kind of "why" comment — future me or a human reviewer sees the whole rationale at the site of the decision instead of having to chase PR comments. Good.

A few concrete notes on the new position:

- **Arithmetic nit in the expanded comment.** The block comment at `styles.css:4582-4584` says *"within the ~30 px populated bar (24 px min-height + 6 px vertical padding + 1 px top border)"*. `padding: 3px 12px` on `.claude-usage-bar` is 3 px top + 3 px bottom = 6 px total, so the arithmetic is correct this time (my previous review's version of this comment had it at "3 px vertical padding" which was wrong). Good fix, just flagging that this version reads correctly.
- **Dot vs wordmark vertical alignment at 7 px.** `display: flex; align-items: center` on `.app-logo` handles the dot-vs-wordmark alignment within the flex container, so the 15 px wordmark and 13 px dot sit on the same baseline relative to each other regardless of the outer `bottom` value. That's unchanged. What *is* new is that the whole flex container now sits nested inside the bar rather than floating above it — if the bar's actual rendered height differs from the comment's ~30 px (e.g. a theme that sets a larger `font-size` on `.claude-usage-bar .usage-item`, or a user-agent with a different default `line-height`), the overlay could peek above or below the band. The `min-height: 24px` locks the floor but not the ceiling. Probably fine in practice; worth knowing.
- **Glow halo at 10 px.** `app-logo-pulse-red` at 50% of its cycle puts a 10 px-radius rgba halo around the dot (`styles.css:4695-4698`). At `bottom: 7px` with a 13 px dot, the dot's top edge sits at 20 px and the halo extends to ~30 px — so the halo's top reaches into the border/divider region of the bar, which is the "minimally overlaying padding" tolerance the owner accepted. Blue halo at 8 px radius is tighter and fully inside the bar. Fine.
- **The block comment carries PR context into the codebase.** The expanded comment explicitly names `@soderbjorn` and the two screenshots. That's unusually personal for a source comment — it locks the explanation to a specific review thread that won't be easy to find in five years. Consider softening to "Earlier iterations at 32 px collided with the bar content; at 44 px the overlay sat inside the tab content surface rather than inside the bar band" without the `@soderbjorn` attribution. Nit.

Kotlin/JS documentation in `AppLogo.kt` and the diff in `WebState.kt` / `ClaudeUsageBar.kt` is unchanged on this head and still conforms to CLAUDE.md (what it does, who calls it, parameters, `@see` cross-references).

Build: I did not run `./gradlew :web:build` on this head. `01ce393` is CSS-only on top of a build-clean head, so I'd expect it to pass; the PR body (updated in the follow-up comment) reports a local build pass.

## Consequences
- Users who previously collapsed the Claude usage bar will hit the `.collapsed { display: none }` interaction described in the Functional review on this head, same as on the previous head — the "always reserve space" invariant from issue #14 still breaks for that population. This is a real behaviour shift for existing users (the bar silently re-appears the first time a null envelope lands) and should be closed out before merge.
- No data-model, schema, or API changes. Web-only. Android/iOS unaffected (web-specific HTML/CSS/Kotlin-JS).
- Electron window-drag behaviour is unchanged — the overlay sits outside `.app-header`, so the no-drag allowlist from the project's memory rule doesn't apply.

## Risks
- **Should-fix:** The four items from my previous reviews (collapsed-vs-empty class overwrite; stale `applyUsageBarCollapsedState` KDoc; empty-vs-populated height jump; hardcoded dot colours). The first one is the only live behaviour regression against the issue's contract; the other three are polish/hygiene. A ~15-line third commit across `ClaudeUsageBar.kt`, `WebState.kt`, and `styles.css` would close all four.
- **Nit:** The `@soderbjorn` attribution in the expanded block comment at `styles.css:4572-4591` ties the source comment to a specific PR thread. Rewording to describe the iterations functionally would make the comment age better.
- **Nit:** Glow halo on `app-logo-pulse-red` at 7 px `bottom` reaches into the bar's top border area. Accepted by the owner per the "only minimally overlaying padding" tolerance, but worth knowing if the bar's border styling ever changes.

No security, performance, or race-condition concerns. CSS-only follow-up; the pulse animations are keyframes rather than `setInterval`, so nothing to leak.

## Ideas & alternatives
If the collapsed-vs-empty interaction is resolved by making `.collapsed` a no-op (since the bar is always-on now), the `usageBarCollapsed` state bit and the `applyUsageBarCollapsedState` function become dead wood. Worth either removing them in the follow-up commit or leaving a `// TODO: dead since #16, keep for rollback` comment to flag future removal.

## Verdict
The position fix in `01ce393` is correct and directly answers the second screenshot — the logo now lives in the bottom bar rather than inside the tab content area. That part is a "merge" for me. The remaining four items (one behaviour regression, three nits) are carryovers from the first two reviews that neither follow-up has picked up, and at least the collapsed-bar interaction is worth closing out before merge so the issue's always-reserve-space contract actually holds for every user. I did not exercise the UI in a browser (sandbox has none), so the owner's visual confirmation on the new 7 px placement remains the authoritative word on how it actually reads.

PR #17, opened 2026-04-24
Inset the Android terminal surface so text clears the screen edges

### LINKED ISSUE #15: Android app missing some terminal padding

soderbjorn:
Closes #15

## Summary
Adds a small Compose-level inset around the Termux `TerminalView` on the Android terminal screen so rendered glyphs no longer sit flush against the screen's left/right edges on phones. The inset is 6 dp horizontal and 2 dp vertical, applied to the `Box` that hosts the `AndroidView` holding `TerminalView`; this flows naturally through `TerminalView.updateSize()` to produce a slightly narrower cols grid without touching any vendored terminal code.

## Background
Issue #15 reports that "text appears sometimes slightly off the edge (or at least slightly too close to the edge) in the Android app while viewing a terminal" and asks for "some (minimal) padding". The attached screenshot shows a phone portrait view with the output column rendering right up against the left edge of the screen. Nothing is actually clipped by the OS — the letterforms just have zero breathing room against the bezel, which reads as cramped.

## Approach
The terminal is rendered by a Termux `TerminalView` embedded via `AndroidView` inside a `Scaffold` `Column` → `Box` in `TerminalScreen.kt`. Previously the `Box` was sized with just `.weight(1f).fillMaxWidth()`, so the `TerminalView` filled it edge-to-edge. The change adds `.padding(horizontal = 6.dp, vertical = 2.dp)` to that `Box`. The inner `AndroidView` still calls `.fillMaxSize()`, so the `TerminalView` gets the inset-reduced area. `TerminalView.updateSize()` reads `getWidth()` / `getHeight()` on every size change and derives new cols/rows from the font metrics — so the grid automatically adapts to the slightly smaller viewport and the server receives the correct resize. No grid-metrics code, no color/theme code, and nothing in the vendored `terminal-view` or `terminal-emulator` modules had to change.

### Padding at the Compose `Box` layer, not as Android view-padding on `TerminalView`
The obvious-looking alternative is `AndroidView(modifier = Modifier.padding(...))` or calling `view.setPadding(...)` on the `TerminalView` in the `factory` lambda. Both would fail silently: `TerminalRenderer` in the vendored `terminal-view` module draws against `view.getWidth()` / `view.getHeight()` and `mRenderer.mFontWidth` — it does not consult `getPaddingLeft()` / `getPaddingTop()` at all (see `TerminalView.java:989-995`). Setting Android view padding would therefore reserve the space but still paint over it, moving the problem rather than fixing it. Shrinking the hosting `Box` at the Compose layer makes the `TerminalView` itself measure smaller, so its internal draw math stays consistent. The trade-off is that the padded band has the surface background colour rather than the terminal background colour — since the hosting `Column` already paints with the resolved terminal `bgComposeColor`, these colours match and the inset is invisible as a seam.

### Asymmetric inset: 6 dp horizontal, 2 dp vertical
The crowding in the issue's screenshot is a horizontal problem: the leftmost column of characters touches the screen edge. Vertical spacing is already comfortable because the `TopAppBar` sits above the terminal and the `ImeHelperToolbar` (plus optional `SwipeInputBar`) sits below — both provide generous vertical separation from the screen bezel. A small 2 dp vertical inset keeps the topmost row / bottommost row from sitting right on the edge of the internal background rectangle while still being effectively invisible. The horizontal value (6 dp) is a judgement call — enough to visibly lift glyphs off the edge on a 1080 px-wide phone (≈18 px at xxhdpi) without sacrificing a whole column on narrow devices. The issue explicitly asks for a "minimal" amount, so I stayed conservative; if it reads too tight on physical hardware the value is a one-line tweak.

### No new user-facing setting for the inset
I considered exposing the inset as a setting (so power users can go back to edge-to-edge), but the issue frames this as a fix rather than a preference, and the inset is small enough that it shouldn't bother anyone. Adding a preference would also expand the `UiSettings` shape for a cosmetic toggle, which runs against the "simple JSON settings" preference recorded in the user's project memory. If demand surfaces, a global terminal-padding knob can be added later as a single numeric value in the settings JSON.

### Scoped to Android (no web / iOS / Electron changes)
The issue is tagged `ai-dev` and describes the Android app specifically. The web front-end already wraps the xterm.js surface in a padded container (the `.pane-inner` / terminal wrapper rules in `styles.css`), and the Electron app reuses the same web UI. iOS uses a different terminal stack entirely. So the fix lives only in `TerminalScreen.kt` — no cross-platform parity work was needed here.

## Assumptions
- The screenshot in the issue represents the general case, not an edge case of a specific theme or column count. If there's a particular device / configuration where the crowding is worse, the 6 dp value may need bumping.
- The surface behind the inset band is the terminal's own background, which is correct because the hosting `Column` already uses `bgComposeColor` from the resolved palette. I verified this in code but not in a running app.

- **`view.setPadding(...)` on the `TerminalView`** — silently ineffective; `TerminalRenderer` ignores view padding and paints over it.
- **Modifying the vendored `terminal-view` renderer to respect padding** — out of scope and the `terminal-view` / `terminal-emulator` modules are explicitly excluded from the project's documentation / modification rules in `CLAUDE.md`. Keeping the fix on the consumer side avoids divergence from the upstream fork.
- **Adding the inset on the outer `Column` (the one that takes `paddingValues`)** — that would also pad the `ImeHelperToolbar` and `SwipeInputBar`, which are already correctly flush; only the terminal surface needs the breathing room.

## Verification
- `./gradlew :androidApp:assembleDebug` passes locally (clean build, no warnings introduced by the change).
- Manual on-device visual verification **not performed** — this sandbox has no Android device / emulator available. The user should sanity-check the result on hardware: (a) the leftmost column of terminal text no longer touches the screen edge, (b) the padded band is not visible as a colour seam (should blend with the terminal background), (c) column count on resize / rotation still matches what the server expects (check by running `stty -a` or observing that Claude Code's UI doesn't reflow incorrectly), and (d) no regression to the pinch-to-zoom gesture (the `TerminalView` still fills the padded `Box`, so gestures work as before).

## Files of note
- `androidApp/src/main/kotlin/se/soderbjorn/termtastic/android/ui/TerminalScreen.kt` — one change: the inner `Box` that hosts the `TerminalView` gains `.padding(horizontal = 6.dp, vertical = 2.dp)`, and an inline comment explains why the inset has to live here rather than on the `AndroidView` or the `TerminalView` itself.

## Follow-ups
- If the inset value turns out to be wrong on real hardware (too tight on large phones, too loose on small tablets), it could be promoted to a density-aware constant or made theme-configurable. Starting with a hard-coded conservative value keeps this PR minimal, in line with the issue's "minimal" framing.

### DIFF
--- a/androidApp/src/main/kotlin/se/soderbjorn/termtastic/android/ui/TerminalScreen.kt
+++ b/androidApp/src/main/kotlin/se/soderbjorn/termtastic/android/ui/TerminalScreen.kt
@@ -491,10 +491,23 @@ fun TerminalScreen(
                 .padding(paddingValues)
                 .background(bgComposeColor),
         ) {
+            // Small inset around the terminal render surface so glyphs in
+            // the leftmost column / topmost row don't sit flush against the
+            // screen edge. Termux's TerminalView draws directly against its
+            // own getWidth()/getHeight() — it ignores Android view padding —
+            // so we apply the inset at the Compose layer by shrinking the
+            // Box that hosts it. The resulting smaller viewport flows back
+            // through TerminalView.updateSize() to produce an appropriate
+            // (slightly smaller) cols/rows grid automatically.
+            // Horizontal inset is larger than vertical because the
+            // screenshot in issue #15 shows horizontal crowding on phones
+            // while the IME toolbar already provides vertical separation
+            // at the bottom and the top app bar does the same at the top.
             Box(
                 modifier = Modifier
                     .weight(1f)
-                    .fillMaxWidth(),
+                    .fillMaxWidth()
+                    .padding(horizontal = 6.dp, vertical = 2.dp),
             ) {
             AndroidView(
                 modifier = Modifier


soderbjorn:

## Summary of changes
Adds a Compose-level inset of 6 dp horizontal and 2 dp vertical to the `Box` that hosts the Termux `TerminalView` on the Android terminal screen, so the rendered glyph grid no longer touches the screen's left/right edges. Because Termux's `TerminalView` draws directly against `getWidth()` / `getHeight()` and ignores Android view-level padding, the inset has to be applied by shrinking the hosting Compose `Box` rather than by setting view padding — shrinking the `Box` propagates into `TerminalView.updateSize()` and produces a slightly narrower cols grid automatically, with the correct resize being pushed back to the server. One file, 14 additions / 1 deletion (the old `fillMaxWidth(),` line becomes `fillMaxWidth()` plus a new `.padding(...)` line plus a long explanatory block comment). The PR body's behavioural description matches the diff exactly.

## Functional review
The fix is well-reasoned and should do what the issue asks. The key claim — that Termux's `TerminalView` ignores Android view padding — is correct: `TerminalView.updateSize()` at `terminal-view/src/main/java/com/termux/view/TerminalView.java:988-1003` reads `getWidth()` / `getHeight()` directly with no `getPaddingLeft()` / `getPaddingTop()` subtraction, and `onDraw()` at `terminal-view/src/main/java/com/termux/view/TerminalView.java:1020-1043` delegates unconditionally to `mRenderer.render(...)` with no clip for padding. So `view.setPadding(...)` would indeed have reserved the band and then painted over it. Applying the inset at the Compose layer is the right place to do it.

The colour match of the inset band checks out: the hosting `Column` at `androidApp/src/main/kotlin/se/soderbjorn/termtastic/android/ui/TerminalScreen.kt:491` paints with `bgComposeColor = Color(terminalPalette.terminal.bg)` (defined at line 235), and the `TerminalView` itself gets `view.setBackgroundColor(bg)` in `applyTerminalColors` at line 786 using the same `palette.terminal.bg.toInt()` source. Both sides of the seam are the same colour, so the inset band reads as part of the terminal surface rather than a frame around it. Good.

A few edge cases the author did not call out that are worth thinking about:

- **Horizontal system-bar insets in landscape.** The outer `Column` at line 490 already applies `paddingValues` from the `Scaffold`, and on Android this typically includes horizontal safe-area insets for three-button nav bars in landscape. The new 6 dp inset is applied *after* that, so in landscape the terminal's usable width is `screenWidth − navBarInset − 12 dp`. On a phone in landscape with a tall text size, this could cost another full character column on each side. Not a regression (the pre-PR code didn't lose that column either, so this is a strict cosmetic loss of two columns) but worth being aware of — some users pack landscape terminals very densely.
- **Cost of the inset at high densities.** `view.setTextSize(30)` at line 537 gives a font width of roughly 18 px at xxhdpi. 6 dp at 3x density is 18 px — close to one character column each side, or ~2 columns total. That's the real cost of the fix in "columns visible". On a 1080 px-wide phone portrait at text size 30, you go from ~60 cols to ~58 cols. Conservative and in line with the issue's "minimal" framing, but it's a real measurable shift, not just cosmetic.
- **Pinch-to-zoom interaction.** The inner `AndroidView` at `TerminalScreen.kt:512` still uses `.fillMaxSize()`, so the `TerminalView` fills the now-smaller `Box`. `onScale` (line 548) changes font size, which triggers `updateSize()` and a new cols/rows negotiation — the padded viewport feeds cleanly into that path with no special casing needed. Pinch behaviour is preserved.
- **Rotation / `onConfigurationChanged`.** Rotation re-runs the Scaffold layout, re-measures the padded `Box`, and invokes `TerminalView.updateSize()` with the new viewport. No state in the grid-size listener at line 529 that would break under a size change — the existing `lastSent` guard already debounces.
- **Not exercised in this review.** I didn't build the Android app (`./gradlew :androidApp:assembleDebug`) and cannot run an emulator in this sandbox, so visual confirmation against the screenshot in #15 is on the author. The PR body's self-assessment — "manual on-device visual verification not performed" — is honest about this and lays out a sensible checklist for when it happens.

Nothing jumped out as a behavioural regression.

## Technical review
The change is a single `.padding(horizontal = 6.dp, vertical = 2.dp)` modifier appended to the existing `Modifier.weight(1f).fillMaxWidth()` chain. Modifier order is correct: `weight` measures the `Box` to fill remaining column space, `fillMaxWidth` fills horizontally, and `padding` is applied inside that measured size (Compose padding reserves space by shrinking the content area). The `AndroidView` inside uses `.fillMaxSize()` (line 513) so it naturally picks up the padded size. No imports needed — `dp` is already imported at `TerminalScreen.kt:84`. No vendored `terminal-view` / `terminal-emulator` modules were touched, which is correct per the project's documented exclusion of those modules from modification rules.

The accompanying block comment at `TerminalScreen.kt:494-503` is substantive — it explains *why* the inset lives here rather than as view padding (the `getWidth()`/`getHeight()` footgun), which is the part a future reader is most likely to mess up. That kind of "why" comment is exactly what CLAUDE.md asks for, and it's placed at the site of the decision rather than in a separate doc. Good. The 12-line comment for a 2-line change is an acceptable ratio given how non-obvious the constraint is.

Naming and placement are fine; nothing else in the file was touched. `changedFiles: 1`, `additions: 14`, `deletions: 1` — minimal and in line with what the PR body promises.

A few nits with specific pointers:

- **Indentation nit.** The `AndroidView(` block at `TerminalScreen.kt:513` is at the same column as the outer `Box` opening brace at line 508, rather than being indented one level further to reflect its nesting inside that `Box`. This predates the PR (the existing code already had this anomaly — `fillMaxWidth(),` sat at column 17 under a `Box` whose opening brace was at column 17), and the PR didn't introduce it, but while adding 12 lines of commentary in the immediate area it's cheap to re-indent the `AndroidView(...)` call four spaces to the right and move the closing `            }` to match. Not a blocker; the compiler doesn't care, but human readers briefly lose the nesting. If you'd rather keep the PR surgically minimal, leave it — it's a pre-existing style issue, not a new one.
- **Literal constants, no named `val`.** `6.dp` and `2.dp` are inline at the call site with the rationale in the comment above. That's fine for a one-off, and extracting `private val TerminalEdgeInsetHorizontal = 6.dp` would be overkill for two numbers used in one place. If the follow-up the PR body hints at ever materialises ("density-aware constant or theme-configurable"), that would be the time to name them.
- **No settings knob.** The PR body explicitly rejects adding a user-facing toggle for the inset, citing the project's "simple JSON settings" preference. I agree with that call — exposing 6 dp / 2 dp as a user setting is overkill for a cosmetic fix, and the JSON blob convention would encourage adding it only once there's demand.
- **No test coverage.** There are no existing tests under `androidApp/` that exercise terminal sizing, so there's nowhere to plug a regression check in. Composable layout tests for this would require instrumentation, which this codebase doesn't currently set up. Not a gap introduced by this PR.

Build: I did not run `./gradlew :androidApp:assembleDebug`. The PR body claims it passes locally.

## Consequences
- **Column count drops by ~2 at typical densities.** As noted in Functional review, 6 dp × 2 at xxhdpi is roughly one character column each side with the default 30sp text size. Users on narrow phones will see slightly less horizontal content. In practice the server sees the smaller grid via the `setOnTerminalGridSizeChangedListener` path (line 529) and `stty cols` will reflect it correctly, so nothing downstream of the PTY breaks — it's purely a viewport reduction.
- **No schema, no settings, no API surface changes.** Purely a layout tweak in one Compose function.
- **Android-only.** Web / Electron / iOS are unaffected; the PR body correctly notes the web terminal already has CSS padding on its wrapper. No cross-platform parity work needed.
- **Trivial to revert.** One-line revert if the value turns out to be wrong on hardware.

## Risks
- **Nit:** The `AndroidView`'s indentation still sits one level too shallow for its new nesting (pre-existing, not introduced by this PR). Cosmetic.
- **Nit:** 6 dp is a magic number. If the value needs tuning on real hardware, a named constant or a follow-up density/theme knob is the cheap next step — which the PR body already flags as a follow-up.
- **Not exercised:** Visual verification on an Android device was not performed in this environment. The PR body acknowledges this explicitly and lists what the author should check on hardware.

No security, performance, or race-condition concerns spotted. The modifier chain is synchronous layout work — no background threads, no new allocations per frame.

## Ideas & alternatives
- If the 6 dp value later turns out to be wrong in some theme or on some form factor, the cleanest knob is probably a single numeric field in the existing `UiSettings` JSON (e.g. `android.terminalHorizontalInsetDp: number`) with a default of 6. That would keep the "simple JSON settings" contract intact and avoid adding typed DTO fields. Not worth doing speculatively.
- For cross-platform consistency, the web terminal's `.pane-inner` padding and the Android inset could eventually be driven off the same semantic token (e.g. `--t-terminal-edge-inset`), though that's a bigger refactor than this PR should carry. Consider if a third platform ever needs the same breathing room.

## Verdict
Merge as-is (with the optional cosmetic indentation touch-up, if you feel like it). The change is small, correct, well-reasoned, and accompanied by the right amount of explanatory comment for a non-obvious constraint (the Termux view's padding-ignoring draw path). The only real unknown is whether 6 dp is the right amount on actual hardware, which is a judgement call the author explicitly flags as tunable. I did not exercise the app visually, so the author's device-verification checklist remains the authoritative word on how it looks on a phone.

PR #40, opened 2026-06-22
Mac: New tab/terminal, directional pane nav, and Quake show/hide shortcuts

mil0m:
## Summary

Adds a set of keyboard shortcuts to the **Mac/Electron app**, aimed at agent-development workflows:

| Shortcut | Action |
| --- | --- |
| **⌘T** | New tab |
| **⌘D** | New terminal pane (in the active tab) |
| **⌃⌥← / ↓ / ↑ / →** (and **⌃⌥H/J/K/L**) | Move pane focus by direction (vim-style), with wrap-around |
| **Ctrl+`** (and Ctrl+Alt+Cmd+Space) | Quake/iTerm-style global show/hide toggle |

## Details

**⌘T / ⌘D** — native **File** menu items in the Electron main process → `new-tab` / `new-terminal` IPC → preload bridge → renderer dispatches the same `WindowCommand.AddTab` / `AddPaneToTab` the `+` buttons use. Purely additive; no protocol/server changes. Both are also surfaced in the ⌘/ cheatsheet.

**Directional pane navigation** (`web/.../PaneNavigation.kt`) — replaces the darkness-toolkit's linear pane *cycle* (which was on the same ⌃⌥-arrow chord) with true directional movement. A `window` capture-phase keydown listener is installed *before* the toolkit shell so it preempts the toolkit's dispatcher (`stopImmediatePropagation`); it matches on `event.code` (Option mutates `event.key` on macOS) and requires exactly Ctrl+Option, so the toolkit's ⌃⌥⇧-arrow **tab** cycle is untouched. The spatial pick reads live `.dt-pane` DOM rects (pane geometry is toolkit-owned, so the server config is unreliable for this). It sends **`SetFocusedPane` only** — matching the cycle it replaces, which was focus-only (no z-order re-stacking). HJKL works but is intentionally left out of the cheatsheet as a power-user binding.

**Quake toggle** (`electron-main/.../ElectronMain.kt`) — the existing global summon (Ctrl+Alt+Cmd+Space) now *toggles* (hides when frontmost) instead of only showing, and **Ctrl+`** is added as a second global hotkey with the same `toggleQuakeWindow()` behavior.

## Caveats worth a reviewer's eye

- **`Ctrl+\`` is a *global* shortcut** — while termtastic runs it shadows Ctrl+` in every app (e.g. VS Code's integrated-terminal toggle). This is inherent to a Quake hotkey window and was a deliberate choice.
- Global hotkeys remain **dev-gated** (`registered only when !IS_DEV_LAUNCH`), so they activate only in packaged builds, not dev/demo launches.
- ⌃⌥-arrows overlap macOS **VoiceOver** navigation when VoiceOver is on (pre-existing; the toolkit already used ⌃⌥-arrows).

## Testing

- ⌘T, ⌘D, directional pane nav (incl. wrap + over a focused terminal): verified in `:electron:runDemo`.
- Quake toggle: verified in a packaged build (the dev launches don't register global hotkeys).
- All modules compile.

## Notes

- Design spec under `docs/superpowers/specs/`.
- An independent code-review pass was run on the branch; its findings were addressed (notably: directional nav switched to focus-only to match the prior cycle, and corrected docs around maximized-pane handling).

### DIFF
--- a/docs/superpowers/specs/2026-06-22-directional-pane-navigation-design.md
+++ b/docs/superpowers/specs/2026-06-22-directional-pane-navigation-design.md
@@ -61,10 +61,12 @@ pane in direction" from the live on-screen rectangles:
 
 - Candidate panes: `document.querySelectorAll(".dt-pane")` (only the
   active tab's panes are in the DOM). Each carries `data-pane-id`.
+  Minimized panes are already absent (rendered as dock items).
 - Current pane: `.dt-pane.dt-pane-focused`, falling back to the active
   tab's `focusedPaneId`, then the first `.dt-pane`.
-- Skip zero-size rects (hidden / `display:none`, e.g. siblings of a
-  maximized pane).
+- Skip only genuinely zero-size rects (defensive, e.g. a pane
+  mid-collapse-animation). NB: a sibling *covered* by a maximized pane
+  keeps its full rect and remains a valid target — see Focus dispatch.
 
 ### Pure selection function (testable, DOM-free)
 
@@ -88,14 +90,21 @@ pickPaneInDirection(rects: List<Rect>, currentIndex: Int,
 
 ## Focus dispatch
 
-On a chosen target pane id, send the same pair a sidebar click uses:
+On a chosen target pane id, send **`SetFocusedPane` only**:
 
 ```
 SetFocusedPane(tabId = activeTabId, paneId = target)
-RaisePane(paneId = target)
 ```
 
 `activeTabId` from `latestWindowConfig?.activeTabId`. No-op if null.
+
+We deliberately do **not** send `RaisePane`. The toolkit's own
+Ctrl+Alt+Left/Right pane cycle (which this replaces) was focus-only and
+never re-stacked z-order — only an explicit pane *click* raises. Matching
+that avoids churning the persisted layout z-order on every keystroke. If
+the target is covered by a maximized sibling, the server's
+`setFocusedPane` clears that sibling's maximize flag, so the focused pane
+becomes visible (mirroring the cycle's auto-unmaximize) without a raise.
 
 ## Gating & cheatsheet
 
--- a/docs/superpowers/specs/2026-06-22-directional-pane-navigation-design.md
+++ b/docs/superpowers/specs/2026-06-22-directional-pane-navigation-design.md
@@ -1,0 +1,138 @@
+# Directional pane navigation (vim-style) — design
+
+**Date:** 2026-06-22
+**Scope:** Mac/Electron app (web frontend, `web/src/jsMain`)
+**Status:** Approved
+
+## Goal
+
+Let the user move keyboard focus between panes in the active tab by
+*direction* (left / down / up / right), vim-style, instead of the
+toolkit's current linear *cycle*.
+
+## Bindings
+
+Directional pane focus, four directions, two key families that behave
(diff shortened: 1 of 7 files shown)


soderbjorn:
@mil0m Thanks for your PR! Some comments:

- First of all, most or all of this should be done in the Darkness toolkit instead: https://github.com/soderbjorn/darkness-toolkit/
- I like the idea of directionally moving between windows. We already support CTRL-OPT-left/right but my understanding is that you'd be added up/down to make it more spatial, which seems good.
- CMD-T / D - good stuff.
- There is support for CTRL-ALT-CMD-space as a global hotkey, is it the "toggling" functionality that is the main attraction of the change, or is it that you specifically want CMD-backtick?

mil0m:
> @mil0m Thanks for your PR! Some comments:
> 
> * First of all, most or all of this should be done in the Darkness toolkit instead: https://github.com/soderbjorn/darkness-toolkit/
> * I like the idea of directionally moving between windows. We already support CTRL-OPT-left/right but my understanding is that you'd be added up/down to make it more spatial, which seems good.
> * CMD-T / D - good stuff.
> * There is support for CTRL-ALT-CMD-space as a global hotkey, is it the "toggling" functionality that is the main attraction of the change, or is it that you specifically want CMD-backtick?

* Seems like the correct move to place this in Darkness! 
* Yes especially when adding a bit more panes it's much easier to jump between them if it can be done in all spatial directions.
* It's the toggling that's the main feature here. CMD-backtick was just my personal flavour before asking for mappable shortcuts

ottojaa:
Chiming in as a user, what I'd love to see: turn StandardHotkeys into a configurable keymap (ActionId → Hotkey, defaults = today's constants), have installTabNavigationHotkeys/LayoutRenderer read from it, and persist overrides through the existing Persister/UiSettings path - with a small rebinding UI in the settings sidebar. Hotkey already serializes trivially and toChordLabel() handles display, so the lift is mostly wiring. 
 
Before switching to Termtastic, I was pretty happy with my ghostty + zellij setup allowing for cmd + arrow keys left / right navigation between tabs, but the configuration was pain because you had to first wire both to talk with each other (i.e decide that cmd+arrow means alt+arrow). But here I think we have a chance to make Termtastic's binding configurability much better than that.
 
Tbh I'd like (almost) everything to be configurable 😁 with sensible defaults of course - I think I'm in the minority of preferring cmd over alt.

Highly agree with @mil0m about the directional navigation!

soderbjorn:
Thanks for the feedback! The next question, then, is if custom hotkeys should be synced with the server, so that the ones that can will also apply for remote web clients. Probably they should be because other appearance settings are.

soderbjorn:
Maybe a good opportunity to play a bit with Mythos now when it's apparently back...

soderbjorn:
I have some stuff in progress for this, will update you soon...

soderbjorn:
I did some stuff related to the hotkeys along your request in the latest version, 1.4:
(see version log at https://github.com/soderbjorn/termtastic/releases)

However, I might have been a bit quick with the testing as not everything looks tip top with the hotkeys still (for one, not all hotkey actions are editable).

soderbjorn:
@mil0m I'll close this as it's so old. Have a look at the latest version when you have time and feel free to open a new PR if it still feels relevant!

PR #61, opened 2026-06-25
Mac: Add Cmd+1..9 to switch tabs by position

mil0m:
## What

Adds **⌘1–⌘9** keyboard shortcuts to jump straight to a tab by position in the Electron desktop app:

- **⌘1–⌘8** activate the 1st–8th tab by raw index in `WindowConfig.tabs`.
- **⌘9** always activates the **last** tab (the Safari / Chrome / iTerm convention).
- Plain ⌘ only — any combination with Ctrl/Alt/Shift is ignored so existing chords aren't shadowed.

## How

- The pure index mapping lives in commonMain: `resolveTabSwitchIndex(digit, tabCount)` (`client/.../TabSwitchShortcuts.kt`), with unit tests.
- `installTabNumberShortcuts()` (`web/.../TabNumberShortcuts.kt`) installs a capture-phase `keydown` handler that matches `⌘+<digit>` via `event.code` (`Digit1`–`Digit9`), resolves it against the live `latestWindowConfig`, and fires `WindowCommand.SetActiveTab`.
- Wired in `bootViaToolkitShell`, gated to `isElectronClient` — a real browser owns `⌘+<digit>` for switching browser tabs, so we don't fight it there.
- Cheatsheet (`⌘/Ctrl+/`) lists the chord on the desktop client.

## Testing

- ✅ `:client:jvmTest` — resolver unit tests (positional mapping, ⌘9→last, out-of-range, empty, invalid digit).
- ✅ `:web:compileKotlinJs` — compiles clean.
- ✅ Manually verified in the Electron demo (`scripts/run-electron-demo.sh`): ⌘1/2/3 switch tabs, ⌘9 jumps to the last tab.

## Note for the maintainer (unrelated pre-existing issue)

`main` currently doesn't compile the web target: commit dd48891 changed `web/.../PaneTypeModal.kt` to reference a `LinkThumbnail` class in `jsMain`, but the file defining it isn't committed (it only exists in the Android module). This PR doesn't touch that — I verified the build locally with a throwaway `LinkThumbnail` shim that is **not** part of this branch. You likely have an uncommitted `LinkThumbnail.kt` to add.

### DIFF
--- a/client/src/commonMain/kotlin/se/soderbjorn/termtastic/client/TabSwitchShortcuts.kt
+++ b/client/src/commonMain/kotlin/se/soderbjorn/termtastic/client/TabSwitchShortcuts.kt
@@ -1,0 +1,39 @@
+/* TabSwitchShortcuts.kt (commonMain)
+ *
+ * Pure mapping logic for the `Cmd+<digit>` tab-switch keyboard
+ * shortcuts (Mac/Electron). Kept platform-agnostic and free of any
+ * DOM / WindowConfig dependency so it can be unit-tested directly; the
+ * web frontend feeds it the live tab count and uses the returned index
+ * to pick the target tab id. See the jsMain `installTabNumberShortcuts`
+ * wiring for the event side of this feature.
+ */
+package se.soderbjorn.termtastic.client
+
+/**
+ * Resolves which tab a `Cmd+<digit>` shortcut should activate.
+ *
+ * Convention (matches Safari / Chrome / iTerm):
+ * - Digits `1`–`8` select the 1st–8th tab by raw position (zero-based
+ *   index `digit - 1`), or nothing if that position doesn't exist.
+ * - Digit `9` always selects the **last** tab, regardless of how many
+ *   tabs there are.
+ *
+ * Called by the jsMain key handler ([installTabNumberShortcuts]) on
+ * every matching keydown; the returned index is looked up against the
+ * live `WindowConfig.tabs` list to obtain the tab id passed to
+ * `WindowCommand.SetActiveTab`.
+ *
+ * @param digit the pressed number key, expected in `1..9`.
+ * @param tabCount the number of tabs currently in the window.
+ * @return the zero-based tab index to activate, or `null` when the
+ *   shortcut should be a no-op (no tabs, out-of-range position, or a
+ *   digit outside `1..9`).
+ */
+fun resolveTabSwitchIndex(digit: Int, tabCount: Int): Int? {
+    if (tabCount <= 0) return null
+    return when (digit) {
+        9 -> tabCount - 1
+        in 1..8 -> (digit - 1).takeIf { it < tabCount }
+        else -> null
+    }
+}
--- a/client/src/jvmTest/kotlin/se/soderbjorn/termtastic/client/TabSwitchShortcutsTest.kt
+++ b/client/src/jvmTest/kotlin/se/soderbjorn/termtastic/client/TabSwitchShortcutsTest.kt
@@ -1,0 +1,47 @@
+/**
+ * Tests for [resolveTabSwitchIndex] — the pure mapping from a
+ * `Cmd+<digit>` keyboard shortcut to the tab index it should activate.
+ *
+ * @see resolveTabSwitchIndex
+ */
+package se.soderbjorn.termtastic.client
+
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertNull
+
+class TabSwitchShortcutsTest {
+
+    @Test
+    fun digitsOneThroughEightMapToZeroBasedIndex() {
+        assertEquals(0, resolveTabSwitchIndex(digit = 1, tabCount = 8))
+        assertEquals(2, resolveTabSwitchIndex(digit = 3, tabCount = 8))
+        assertEquals(7, resolveTabSwitchIndex(digit = 8, tabCount = 8))
+    }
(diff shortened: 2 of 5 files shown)


soderbjorn:
This is a good one, but should be done in the Darkness toolkit instead: https://github.com/soderbjorn/darkness-toolkit/issues

soderbjorn:
## Moved this into darkness-toolkit instead 🎁

Per the suggestion that this belongs lower down so it benefits every darkness app (notegrow et al.), I reimplemented the feature in the **darkness-toolkit** web shell rather than in termtastic's `web` module.

**Toolkit commit:** soderbjorn/darkness-toolkit@6b24a7f (on `main`)

### What landed in the toolkit
The shell already owns the tab strip, the tab model (`TabListSnapshot` / `onSelect`), the central `HotkeyRegistry`, and an `isElectron` notion — so positional tab-switching is a natural shell-level concern. Changes (all in `toolkit-web`):

- **`hotkey/StandardHotkeys.kt`** — `tabSwitchHotkey(position)`: a platform-aware chord (Cmd on macOS, **Ctrl on Windows/Linux** — the OS-native convention), alongside the existing `NextTab`/`PreviousTab` constants.
- **`shell/TabBar.kt`** — `renderTabBar`'s `installTabNavigationHotkeys` now also binds 1..9 via the shared `HotkeyRegistry` (same capture-phase listener, replace-on-register, closes over the latest spec). The pure `resolveTabSwitchIndex(digit, tabCount)` mapping is ported verbatim (1–8 positional, **9 = last tab**) and unit-tested in `TabSwitchTest.kt` (the same 5 cases from this PR, now green in `:toolkit-web:jsTest`).
- **`hotkey/Hotkey.kt`** — exposed `isMacPlatform()` and added `isElectronPlatform()` (UA `Electron` token + `process.versions.electron` fallback).
- **`hotkey/HotkeyCheatsheet.kt`** — `tabSwitchHotkeyEntry()`: a ready-made cheat-sheet row so apps advertise the chord with one consistent label instead of hand-writing it.

### Deliberate differences from this PR
- **Resolves against *visible* (non-hidden) tabs**, matching the toolkit's existing `cycleTab` — positions line up with what the user actually sees in the strip, and an already-active target is a no-op (avoids a redundant `onSelect`/server round-trip).
- **Electron gating is runtime-detected** (`isElectronPlatform()`) inside the toolkit, so apps get the correct behaviour automatically with no extra wiring — no per-app boot call to gate.
- **`event.key`-based** (via the toolkit's existing `Hotkey` model) rather than `event.code`. This keeps everything in the toolkit's single hotkey abstraction (one listener, cheat-sheet integration). The tradeoff is layout-independence on e.g. AZERTY; if that matters we can enhance the `Hotkey` model to optionally match by `code`, which would benefit all toolkit chords — happy to do that as a follow-up.

### Suggested path for this PR
Once the toolkit bump flows into termtastic, #61 can shrink to roughly:
1. Bump the `darkness` version and refresh `libs-repo` (`./gradlew publishAllToLibsRepo`) — the behaviour then comes for free.
2. In `TermtasticHotkeysContent.kt`, replace the hand-written entry with `tabSwitchHotkeyEntry()` (still gated on `isElectronClient`).
3. Drop the now-redundant `client/.../TabSwitchShortcuts.kt` (+ its `jvmTest`), `web/.../TabNumberShortcuts.kt`, and the `bootViaToolkitShell` wiring.

Want me to push that slim-down to this branch, or would you rather close #61 in favour of the toolkit change + a follow-up bump?

(The unrelated pre-existing `web` build note about `LinkThumbnail` in your PR description is untouched by any of this.)

soderbjorn:
## ✅ Superseded — this is now on `main`, no action needed here

The feature has shipped via the toolkit and termtastic now consumes it, so **this PR is no longer needed and can be closed.**

**termtastic** `main` → soderbjorn/termtastic@6889352
- Bumps Darkness toolkit `0.2.7` → `0.2.8` and refreshes the committed `libs-repo`.
- Advertises the chord in the cheatsheet via the toolkit's `tabSwitchHotkeyEntry()`, gated on `isElectronClient`.

**darkness-toolkit** `main`
- soderbjorn/darkness-toolkit@6b24a7f — Cmd/Ctrl+1..9 positional tab-switch in the web shell (bound in `renderTabBar`, Electron-gated, resolves against visible tabs, `9 = last`, `resolveTabSwitchIndex` unit-tested).
- soderbjorn/darkness-toolkit@88fd4d9 — version bump to `0.2.8`.

Net effect: the behaviour is now automatic for **every** app that mounts the toolkit shell (termtastic, notegrow, …), not just termtastic.

### Verified
- `:toolkit-web:jsTest` — the 5 resolver cases from this PR, green.
- `:web:compileKotlinJs -Pdarkness.toolkit.useArtifacts=true` — termtastic compiles against the **published** `0.2.8` artifacts (i.e. the no-checkout / CI path works), with the new cheatsheet entry.

### Note
The pre-existing `LinkThumbnail` web-build breakage you flagged in the description is **no longer present** on current `main` — that artifact build above compiled clean.

Thanks for the original implementation — it transferred almost verbatim (the pure `resolveTabSwitchIndex` mapping and its tests moved over as-is). The only behavioural refinements when lifting it to the toolkit: positions resolve against *visible* tabs (consistent with the existing tab-cycle chords) and the modifier is platform-aware (Cmd on macOS, Ctrl on Windows/Linux). Closing in favour of the above — shout if you'd prefer I leave it open for reference.

PR #76, opened 2026-06-30
Mac/Web: Keep scroll position pinned when tailing a full-scrollback live feed

### LINKED ISSUE #31: After scrolling up, text jumps when new content emitted
Reported by Anders J:

"annan kommentar: om man skrollat upp för att läsa saker samtidigt som claude spyr ut sig mer så hoppar skrollpositionen dvs det är nog bara någon offset från bottom istället för relativt en viss rad. Man skulle vilja att bufferten bara växer nedåt (så man måste skrolla mer för att komma ned till botten) istället för att de man läser skiftar"

On my machine I do not see this problem but perhaps you can figure out why this would happen on his machine.

ottojaa:
Closes #31.

## Problem

When tailing a live feed (continuously streaming output) and scrolled up to read back, the viewport keeps drifting upward instead of holding still — the line you're reading scrolls away on every write, making it hard to actually read live logs. Terminals like Ghostty hold the rendered view steady once you scroll up; termtastic doesn't.

This is the bug reported in #31 ("After scrolling up, text jumps when new content emitted"). It also explains the note there that it couldn't be reproduced on the maintainer's machine: the drift only begins **once the scrollback buffer fills** (xterm's default 1000 lines) and trimming starts — a short session never fills it, so it never drifts. Anders J, tailing long Claude Code output, fills it in seconds; a quick local test won't.

## Root cause

`writeHoldingScroll` (`web/.../TerminalHelpers.kt`) re-anchors the viewport to the **absolute** top line (`viewportY`) after each write. That's only correct while the scrollback is still growing.

Once the xterm scrollback fills, xterm trims the oldest line on every write and **decrements `ydisp` itself** to keep the viewed line stationary (verified in xterm 5.3.0 `BufferService.scroll`: `isFull ? isUserScrolling && (ydisp = max(ydisp-1,0)) : (ybase++, isUserScrolling || ydisp++)`). The re-anchor to the now-stale absolute `viewportYBefore` cancels that compensation out and pushes the view up one line per write — the drift.

## Fix

Only re-anchor while the scrollback is still **growing** (`baseY` increased across the write → no trimming, so the absolute anchor is still valid and the existing "snap back if a write followed the bottom" behavior is preserved). Once the scrollback is **full** (`baseY` unchanged), trust xterm's own per-trim `ydisp` compensation and leave the scroll position alone.

```kotlin
if (baseYAfter > baseYBefore) {            // still growing — safe to re-anchor
    val delta = viewportYBefore - viewportYAfter
    if (delta != 0) term.scrollLines(delta)
}                                          // full — xterm already holds the line
```

Checked against all four regimes (growing/full × xterm-held/xterm-followed); the new branch is correct in each, and the growing+followed path still re-sets `isUserScrolling` so subsequent writes stay pinned.

## Scope

Web/Electron (xterm.js) frontend only. No change to the at-bottom auto-follow path.

## Testing

- `./gradlew :web:compileKotlinJs` builds cleanly.
- Manual (`scripts/run-electron-dev.sh`): fill the scrollback past 1000 lines, then stream continuously and scroll up —

  ```bash
  for i in $(seq 1 1500); do echo "warmup $i"; done
  i=0; while true; do echo "log line $i  $(date +%T.%N)"; i=$((i+1)); sleep 0.1; done
  ```

  Before: the viewed lines keep creeping upward. After: the view stays put while new output accumulates below and the "New output ↓" pill appears.

### DIFF
--- a/web/src/jsMain/kotlin/se/soderbjorn/termtastic/TerminalHelpers.kt
+++ b/web/src/jsMain/kotlin/se/soderbjorn/termtastic/TerminalHelpers.kt
@@ -354,19 +354,32 @@ fun markScrollButtonNewOutput(entry: TerminalEntry) {
  * content line* when the user has scrolled up (auto-scroll "pause").
  *
  * The pause must keep what the user is reading completely still — not merely
- * stop short of the bottom. The anchor is therefore the absolute top line
- * (`viewportY`), NOT the distance from the bottom: if we held the distance,
- * every appended line would push the read line up by one and the content would
- * drift past the reader every time output arrives.
+ * stop short of the bottom. xterm.js already does this on its own while its
+ * internal `isUserScrolling` flag is set, and it does so correctly in **both**
+ * scrollback regimes (see `BufferService.scroll`):
  *
- * xterm.js's own `BufferService` already holds `ydisp` when its internal
- * `isUserScrolling` flag is set, so normally a write keeps the line still on
- * its own. We re-assert it defensively in the write callback (after the data
- * is parsed): if anything caused the viewport to follow the bottom, scrolling
- * back up to `viewportYBefore` restores the line *and* re-sets
- * `isUserScrolling` (a negative `scrollLines` sets it), so subsequent writes
- * stay pinned. When xterm already held the line, `delta` is 0 and this is a
- * no-op.
+ *  - **Growing** (scrollback not yet full): each appended line increments
+ *    `ybase` but leaves `ydisp` put, so the read line keeps its absolute index
+ *    and stays on screen.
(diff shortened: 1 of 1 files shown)


soderbjorn:
Thanks for the contribution @ottojaa ! I can't reproduce the problem on my machine for some reason, but I take it that the PR is tested and solves the problem on your computer?

ottojaa:
@soderbjorn Ah yeah the initial problem description is a bit off - it should've mentioned that the issue occurred only when the scrollback buffer was full (started occurring at ~1k lines which made looking at server logs quite difficult). Tested the PR on my machine, fixed the issue.

PR #98, opened 2026-07-02
Honor program-set terminal titles (OSC 0/2) in pane names

ottojaa:
## Summary

Opt-in support for the standard **OSC 0/2 window-title escape sequence**: when a program running in a terminal sets its title, the pane/tab shows it — the same behavior iTerm2, Terminal.app, and ghostty ship by default. Off by default under **App Settings → Experimental features → "Use program-set terminal titles"**.

Title precedence is **manual rename › program title › cwd › fallback**: a pane you renamed yourself is never touched, and turning the toggle off reverts every tab to its folder name.

## Why

Claude Code sets the terminal title to a live summary of the current task (on by default in the CLI). With this toggle enabled, a tab running Claude Code reads like **"Fix login bug"** instead of the working directory — with zero agent-specific integration. The same mechanism covers ssh, vim, and anything else that sets titles.

## How it works

Nearly all the plumbing existed already — titles ride the same pipeline as cwd updates:

- **Capture** — the OSC 7 cwd scanner (`Osc7Scanner`) is generalized to `OscScanner` (git rename, history preserved). It additionally parses `ESC ]0;title ST` / `ESC ]2;title ST` into a per-session `programTitle` StateFlow; OSC 1 (icon-only) is ignored, and sequences split across PTY reads still parse.
- **Gate** — the toggle is a live flow; each session's watcher collects titles through `flatMapLatest(enabled)`. While off, the pipeline is inert. On enable, the program's *current* title applies immediately (StateFlow replay) — no restart, no waiting for the next title change. On disable, stored titles are swept so tabs revert to folder names.
- **Store + resolve** — `LeafNode.programTitle`, persisted like `cwd`; `computeLeafTitle` implements the precedence, and an empty title (the standard "reset") clears back to the cwd name. No client or protocol changes on any platform (`ignoreUnknownKeys` keeps old clients compatible).
- **Sanitize** — titles are untrusted (any program can emit them, including remote hosts over ssh) and are persisted + broadcast to every client. `sanitizeProgramTitle` strips C0/C1 controls, invisible format characters (bidi overrides, zero-widths), box-drawing and HTML-significant characters (defense-in-depth — tab/pane labels render via `textContent` throughout the toolkit), and caps length surrogate-safely.

## Approaches tried and discarded

Two agent-integration designs were prototyped first: screen-scraping the agent's TUI for the user's prompt, and Claude Code hook registration (a loopback HTTP endpoint + a shell wrapper function + a headless `claude -p` call to summarize the prompt into a label). Both worked, but each was ~1.5k lines of externally-coupled, agent-specific machinery with real robustness and privacy questions — prompt text leaving the machine, TUI-format fragility, per-CLI coupling. Plain OSC title support is half the size (+734/−66, of which 288 lines are tests), has none of that surface, and covers every title-setting program instead of one agent.

## Testing

- `ProgramTitleTest` (26 unit tests): sanitizer (HTML/control/C1/bidi stripping, surrogate-safe cap, empty→null), title precedence, apply/clear semantics (manual-rename immunity, multi-pane sessions, no-op paths), the disable sweep, and scanner capture (OSC 0/2, BEL and ST terminators, UTF-8 titles, split sequences, OSC 7 non-interference).
- **End-to-end against a running server**, driven exactly like a client over the `/window` + `/pty/{id}` WebSockets and the settings REST endpoint — 10/10 checks: off → OSC ignored; enable → current title applies immediately; manual rename wins while the program title is recorded underneath; clearing the manual name reveals it; empty title reverts to the cwd name; `<script>`-style titles arrive sanitized; disable → every tab reverts; OSC stays ignored after disable.
  - One design note surfaced by the e2e: sessions restored at startup bind their title watchers before `main()` finishes wiring, so the enable flag is a stable `MutableStateFlow` that `main()` *mutates* — never a replaced reference, which would leave restored sessions deaf to the toggle.

## Open questions

- Is the 750 ms debounce (matching the cwd watcher) the right coalescing for title churn?
- Worth making this default-on eventually, given it's standard terminal behavior elsewhere?

*Demo video in the comments.*

### DIFF
--- a/clientServer/src/commonMain/kotlin/se/soderbjorn/termtastic/Constants.kt
+++ b/clientServer/src/commonMain/kotlin/se/soderbjorn/termtastic/Constants.kt
@@ -1,6 +1,7 @@
 /**
- * Shared network constants used by both the Ktor server and all client targets
- * (Electron, Android, iOS) to agree on connection endpoints.
+ * Shared constants used by both the Ktor server and all client targets
+ * (Electron, Android, iOS) to agree on connection endpoints and on
+ * cross-module UI-settings keys.
  */
 package se.soderbjorn.termtastic
 
@@ -17,3 +18,12 @@ package se.soderbjorn.termtastic
  * connect never triggers a `certificate-error`.
  */
 const val SERVER_TLS_PORT = 8443
+
+/**
+ * UI-settings key (in `/api/ui-settings`) for the opt-in "use program-set
+ * terminal titles" toggle. Defined in the shared module because the key is a
+ * cross-module contract: the web client's App Settings sidebar writes it
+ * (`AppSettingsContent.kt`) and the server reads it live to gate applying
+ * OSC 0/2 program titles to panes (see `TerminalSessions.programTitlesEnabled`).
+ */
+const val TERMINAL_PROGRAM_TITLE_KEY = "terminalProgramTitle"
--- a/clientServer/src/commonMain/kotlin/se/soderbjorn/termtastic/WindowConfig.kt
+++ b/clientServer/src/commonMain/kotlin/se/soderbjorn/termtastic/WindowConfig.kt
@@ -156,13 +156,21 @@ data class LeafNode(
     val sessionId: String,
     /**
      * Display title shown in the pane header. Always equals
-     * `customName ?: prettifyPath(cwd) ?: defaultLabel` — kept denormalized so
-     * the websocket payload doesn't have to compute it client-side and so
-     * persisted blobs round-trip without recomputation.
+     * `customName ?: programTitle ?: prettifyPath(cwd) ?: defaultLabel` — kept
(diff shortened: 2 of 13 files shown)


ottojaa:
https://github.com/user-attachments/assets/34ec18e9-771a-441a-b2fa-459b508bdbf4

In action footage - I think it's a genuinely useful feature as I usually have ~5 terminals for the same project directory, and renaming them constantly when tasks and contexts change is arduous and prone to human errors. WDYT?

soderbjorn:
Thanks! I have not thought about it before, but based on the video, it looks useful. And the fact it's gated behind a setting means that there's no negative consequence of introducing it. So let's merge it!

soderbjorn:
@ottojaa I love this change! So much better!

ottojaa:
@soderbjorn Glad you like it too! 😌

PR #130, opened 2026-07-14
Fix mangled scrollback replay: sanitize queries, persist width, reset on reconnect

ottojaa:
## Problem

Scrollback replay (reconnect and cross-restart restore) had four compounding defects:

1. **Query-reply injection** — recorded PTY output keeps terminal *query* sequences (DSR/CPR `ESC[6n`, DA, OSC 10/11/12 color reads, XTGETTCAP, DECRQSS…). On replay, xterm.js answers them again and the answers are delivered to the live shell as phantom keystrokes — this is the mystery `2cfe2cfe`-style garbage on the prompt (fragments of `rgb:…` color replies).
2. **Width-blind restore** — `pane_scrollback` stored no cols/rows, so restored raw bytes replayed into the default 120×32 grid and then reflowed to the real pane. TUI absolute-cursor repaints landed on the wrong cells → ghost frames, duplicated tables, permanent hard-wrap scars.
3. **Reconnect duplication (web)** — every reconnect re-replayed the full snapshot into an un-reset grid, appending a second copy of the transcript (the native client already prepends `ESC c`).
4. **Handler leak (web)** — `term.onData`/`term.onResize` were re-registered per connection with no unsubscribe, so after N reconnects every keystroke/resize vote was sent N times (multiplying the injected garbage too).

## Fix

- **New `ReplaySanitizer`** (server): one-shot byte filter that strips *complete* query sequences on the replay paths only — `snapshot()` (covers attach replay + persistence, since the saver persists `snapshot()`) and restored-blob ingestion in the session init (cleans blobs written by older versions). The live output stream is untouched, so a running program's query still reaches the client and gets its answer. Lookalikes survive (DECSCUSR `ESC[4 q`, DECSCL, sixel DCS, OSC sets, DECRST/DECSET, SGR); incomplete trailing sequences are preserved because the live stream completes them for a program that is genuinely waiting.
- **Width persistence**: `pane_scrollback` gains nullable `cols`/`rows` (idempotent `ALTER`s migrate existing DBs, same no-`.sqm` pattern as the table's own bootstrap). The saver samples `sizeEvents` with each snapshot; the restore path seeds the fresh session's PTY, headless emulator and `sizeEvents` with the persisted size — the existing Size-then-snapshot protocol then makes clients render the replay at native width before one clean reflow to the live pane. Legacy rows fall back to 120×32 (exact current behavior).
- **Web reconnect reset**: the first binary frame of a connection is the snapshot (protocol sends Size → snapshot → live output); on reconnects the terminal is RIS-reset before writing it — parity with `RealPtySocket`. Input is gated while the replay parses, as belt-and-braces against query families the sanitizer doesn't know.
- **One-time handlers**: `onData`/`onResize` now register once per xterm instance in `ensureTerminal` and resolve the current socket at call time (`TerminalEntry.sendInput` slot / `entry.socket`); demo mode routes through the same single registration.

## Verification

- `:server:test`: 175 tests pass, including 16 new `ReplaySanitizer` cases (all query families, lookalike survivals, ring-boundary rules, idempotence) and 5 new persistence/migration cases (legacy 3-column DB migrates in place and loads with null size).
- End-to-end against a live server with a WebSocket probe: live output still carries raw queries (pass-through intact); reconnect snapshot keeps content but is query-free; after graceful restart the DB holds the blob at 87×41 and the restored session announces 87×41 **before** the replay.
- Also verified the column migration in place on a pre-existing dev database.

### DIFF
--- a/client/src/commonMain/kotlin/se/soderbjorn/lunamux/client/RealPtySocket.kt
+++ b/client/src/commonMain/kotlin/se/soderbjorn/lunamux/client/RealPtySocket.kt
@@ -47,7 +47,7 @@ import se.soderbjorn.lunamux.PtyServerMessage
 /**
  * Live WebSocket connection to `/pty/{sessionId}`. See [PtySocket] for the
  * consumer contract; this class adds the Ktor WebSocket transport, including
- * the server's 64 KB ring-buffer replay as the first frames on [output] and
+ * the server's 128 kB ring-buffer replay as the first frames on [output] and
  * reconnect-with-reset on connection loss.
  */
 class RealPtySocket internal constructor(
--- a/server/src/main/kotlin/se/soderbjorn/lunamux/ServerInitializer.kt
+++ b/server/src/main/kotlin/se/soderbjorn/lunamux/ServerInitializer.kt
@@ -48,7 +48,10 @@ class ScrollbackSaver internal constructor(
     /**
      * Iterate over every leaf in the live config and persist its ring buffer
      * snapshot when [force] is `true` or when the leaf's [TerminalSession.bytesWritten]
-     * has advanced since the previous save.
+     * has advanced since the previous save. The session's effective grid size
+     * is sampled alongside the snapshot so a restore can replay the bytes at
+     * the width they were rendered for (raw PTY output only reconstructs
+     * correctly in a same-width grid).
      */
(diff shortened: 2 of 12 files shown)


soderbjorn:
@ottojaa merged to `main` now.

PR #131, opened 2026-07-14
Replace min() PTY sizing with latest-active-client arbitration

ottojaa:
> **Stacked on #130** (replay correctness) — the first commit here is that PR; review the second commit for this change. Rebases cleanly once #130 merges.

## Problem

A shared PTY is sized to the tiered min() across all attached clients, so:

- a second window, a reconnecting probe, or a paired phone shrinks a session the user is actively typing in — and a dead-but-not-yet-evicted peer pins it small for up to the 40 s ping eviction, with Reformat only helping until the peer re-votes;
- mid-drag transient widths reach the PTY through the 50 ms resize debounce; programs hard-wrap at those widths and xterm.js cannot reflow hard lines → permanent half-width scars in scrollback, worst on split-bar drags (both neighbours get briefly-very-narrow widths).

## Fix

**Server — `ClientSizeArbiter`** (tmux `window-size latest` semantics): the client the user most recently *used* — typed on, reformatted, or opened on a phone — governs the grid. Ambient `NORMAL`/`THREE_D` votes (ResizeObserver refits, window resizes, font loads) update the stored vote without stealing governance. Existing tier semantics are preserved:

- `THREE_D` still overrides 2D viewers (even typing ones) — `Pane.grid3d` behaviour unchanged;
- a `MOBILE` vote still claims the size the moment a phone attaches (readable immediately, your original design), but a desktop keystroke now reclaims it instantly once the phone is idle;
- **no recorded activity → the classic tiered min() decides, bit-for-bit** (pinned by the untouched `Grid3dOverrideTest`), so sessions where nobody types behave exactly as today.

Input frames mark activity via a new `TermSession.noteClientInput` default method — no wire/protocol change, old clients unaffected; `AgentSession` needs no changes. Deliberately not inside `write()`, which MCP tools call with no client identity.

**Web — mid-drag suppression**: automatic PTY votes are gated while a pane-resize gesture is active (capture-phase `pointerdown` on the toolkit's `.dt-pane-separator` / `.dt-pane-corner-resize`, cleared on `pointerup`/`pointercancel`/window `blur` with a final flush). Local refits continue for visual feedback; only PTY sends are held. The debounce rises 50 → 200 ms as the fallback for gesture-less paths (OS window-edge drags). The settle-time `onGeometryChanged` → `forceReassert` is untouched and bypasses the gate.

## Verification

- 19 new `ClientSizeArbiterTest` unit tests + full suite (194 green), including upstream-parity cases (min-within-tier, 3D-beats-normal, mobile-beats-3D with no activity).
- Live WebSocket probe with two desktop clients + an Android-typed client on one session, all 11 scenarios green: ambient attach doesn't shrink; typing transfers governance both ways; mobile claims on attach; desktop keystroke reclaims from an idle phone; phone disconnect while desktop governs is a no-op; Reformat holds against ambient re-votes but yields to typing; the governor's own resize applies immediately.

### DIFF
--- a/client/src/commonMain/kotlin/se/soderbjorn/lunamux/client/RealPtySocket.kt
+++ b/client/src/commonMain/kotlin/se/soderbjorn/lunamux/client/RealPtySocket.kt
@@ -47,7 +47,7 @@ import se.soderbjorn.lunamux.PtyServerMessage
 /**
  * Live WebSocket connection to `/pty/{sessionId}`. See [PtySocket] for the
  * consumer contract; this class adds the Ktor WebSocket transport, including
- * the server's 64 KB ring-buffer replay as the first frames on [output] and
+ * the server's 128 kB ring-buffer replay as the first frames on [output] and
  * reconnect-with-reset on connection loss.
  */
 class RealPtySocket internal constructor(
--- a/clientServer/src/commonMain/kotlin/se/soderbjorn/lunamux/WindowCommand.kt
+++ b/clientServer/src/commonMain/kotlin/se/soderbjorn/lunamux/WindowCommand.kt
@@ -829,11 +829,14 @@ sealed class PtyControl {
      * left alone.
      *
      * The [priority] tier decides how this vote competes with other clients'
-     * (see [SizePriority]): a [SizePriority.NORMAL] vote is the classic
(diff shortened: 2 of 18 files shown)


soderbjorn:
@ottojaa merged to `main` now.

PR #140, opened 2026-07-15
Fix split-pane restore mangling: settle geometry before voting size

ottojaa:
Follow-up on the restore path, on top of `5a670c1` (per-buffer replay ring). This one is entirely client-side.

## Symptom
A terminal in a **split-pane tab** comes back mangled after quit→reopen — cursor-positioned TUI output (Claude Code, `top`, `vim`) scrambled with overlapping columns — while a **single full-window terminal restores perfectly**.

## Root cause
On a cold-restore first attach the client fits xterm to its container in `connectPane`'s `onopen` and casts a soft size vote carrying that fitted width `W_c`. The vote debounces ~200 ms and fires *after* the server has already restored the pane to its persisted width `W_p` and the snapshot has rendered at `W_p`. In a split pane the pane geometry and webfont metrics are still settling at that instant, so `W_c` lands a column or two off `W_p`; the arbiter rebroadcasts it and `applyServerSize` → `term.resize(W_c)` reflows the just-replayed transcript. xterm's reflow is lossy for cursor-positioned output, so the scramble is permanent. A single full-window pane always fits `W_c == W_p`, so the re-vote is a no-op and it never reflows — which is why only multi-pane tabs show it.

## Fix — settle-then-vote
New `TerminalEntry.restoreSettling`, set on first attach. While set, the automatic startup fit/vote paths are held: `sendResize` and `forceReassert` early-return, and the `onopen` / `fonts.ready` / `ResizeObserver` / mount-fit paths re-arm a debounce instead of fitting. `scheduleRestoreSettle` / `finishRestoreSettle` then run a single reconciling fit + soft vote once the container has been quiet for ~250 ms and the webfont has loaded (capped at ~3 s so a stuck font can't strand the pane). 

- Layout unchanged from the pre-quit session → settled fit equals `W_p` → **no reflow**.
- Genuine window-size change → **exactly one** reflow, to the correct width.
- Panes restored into inactive tabs settle on first activation.

`forceReassert` self-guards on socket-OPEN, so a `fonts.ready` firing before `onopen` is a harmless no-op.

## Testing
`:web:compileKotlinJs` clean. This is DOM/xterm sizing with no server-side test harness, so verified by hand: 2-pane tab, Claude Code in one pane, quit → reopen — transcript comes back intact, matching the single-pane case. Confirmed against the `5a670c1` restore path (dead TUI frame rendered as inert styled text below the scrollback).

### DIFF
--- a/web/src/jsMain/kotlin/se/soderbjorn/lunamux/LayoutBuilder.kt
+++ b/web/src/jsMain/kotlin/se/soderbjorn/lunamux/LayoutBuilder.kt
@@ -84,6 +84,15 @@ fun attachDragDrop(container: HTMLElement, term: Terminal) {
  */
 fun sendResize(entry: TerminalEntry) {
     if (entry.applyingServerSize) return
+    // Cold-restore settling window: swallow the vote. The stale fit sampled at
+    // socket open (before the split geometry / webfont settled) debounces
+    // through here ~200 ms later — after the server has already restored the
+    // pane to its persisted width and the snapshot has rendered at it. Voting
+    // that transient width would make the arbiter rebroadcast it and xterm
+    // reflow the transcript to the wrong width (lossy for TUI output). The
+    // single [finishRestoreSettle] pass casts the one authoritative vote once
+    // the pane is stable. See [TerminalEntry.restoreSettling].
+    if (entry.restoreSettling) return
     // Per-pane "stop automatic reflow": when off, never push a resize to
     // the PTY automatically. This is the single chokepoint every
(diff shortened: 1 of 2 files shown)


ottojaa:
Should be the last follow up for the next while! Previously whenever multi-terminal tabs were restored, this is how it looked for me:

Screenshot 2026-07-15 at 19 03 41


Now replays nicely:


Screenshot 2026-07-15 at 19 06 21

soderbjorn:
Looks like working so far. Tricky area. Let's keep our fingers crossed.

PR #146, opened 2026-07-16
Electron auto-updater + local metadata-safe release/publish

ottojaa:
## What

In-app auto-update for the Electron desktop build (`electron-updater`) plus a local, metadata-safe release/publish flow (MacOS).

## In-app updater

- **Main process:** `ElectronUpdaterExternals.kt` + `AutoUpdater.kt` — configure electron-updater, forward lifecycle events → renderer, and check/download/quit-and-install. Guarded on `app.isPackaged`; `autoDownload=false` and `autoInstallOnAppQuit=false` so it's fully **opt-in** (a downloaded update applies only when the user clicks "Restart to install", never silently on quit). quit-and-install stops the bundled server first (`POST /admin/shutdown`) so the relaunched app doesn't reconnect to the stale server. Wired into `ElectronMain.kt` (IPC channels + a "Check for Updates…" Help-menu item).
- **Preload:** `checkForUpdates` / `downloadUpdate` / `quitAndInstall` + `onUpdate*` subscriptions.
- **Renderer:** a compact update **banner pinned to the top of the left sidebar footer** (`AutoUpdaterPanel.kt`, `.lunamux-update-banner`), styled to the footer idiom (theme tokens) — accent title + version, a **"What's new"** link (opens the GitHub release page), a progress bar, and a **Download / Restart to install** action (full-width once downloaded). Clicks transition optimistically so they feel instant. The renderer fires the initial check itself *after* subscribing, so `update-available` can't be missed on launch. A silent **4-hour periodic re-check** keeps long-running sessions informed (skipped mid-flow so a DOWNLOADED banner is never downgraded). The "Connection lost" modal is suppressed during the intentional install teardown.
- Desktop suppresses the `versions.json` update branch (`enableUpdateCheck=false`; **mobile unchanged**); the top-bar bell stays news-only.

## Failure-path hardening

The install teardown assumes nothing:

- `update:quit-and-install` is guarded on a main-side `updateDownloaded` flag, so a state-desynced renderer can't shut the server down for an install that would throw; the install itself is wrapped so a failure surfaces on the banner's error channel and re-arms the quit-confirmation gate instead of dying as an uncaught main-process exception.
- The banner re-enables its action button on every render, so a failed install's "Try again" is actually clickable.
- The "Connection lost" suppression un-sticks when the teardown it anticipated is abandoned: a `quit-cancelled` IPC fires when a kill-server quit is cancelled after a failed shutdown, and the banner's error handler clears it after a failed install — both re-evaluate the aggregate connection status immediately.

## Release / publish (local, metadata-safe)

- `electron/package.json`: mac `zip` target (required for `latest-mac.yml`), GitHub `publish` (draft) provider, `electron-updater` dep, and `dist` never auto-publishes.
- `electron/build.gradle.kts`: **`chmod -R u+w` the jlink'd JRE** — its `legal/` files are read-only, which blocked Squirrel.Mac from stripping the download quarantine on install (the update aborted and relaunched the OLD version; this would have broken production auto-updates too).
- `scripts/release-electron.sh`: builds **locally** (no CI), staples the DMG, and publishes a GitHub **draft** release uploading a fixed glob (dmg + zip + `latest-mac.yml` + `.blockmap`) so metadata can't be forgotten. Flags: `--repo`, `--identity` (self-signed fork testing), `--no-notarize`, `--no-publish` (local build→install loops). Guardrails: test builds **refuse to publish to the production repo**, already-**published** releases are never clobbered, every artifact class is asserted present before upload, and `latest-mac.yml`'s dmg hash is refreshed after stapling rewrites the DMG.

## Verified

- `:electron-main` + `:web` compile.
- **Full end-to-end loop tested** with a free self-signed cert against the fork: banner → Download → progress → Restart → relaunch into the new version. Signature-match + ShipIt install confirmed working (after the JRE-writable fix).
- The release-script guards (production-repo refusal, `--help`) and the `latest-mac.yml` refresh logic exercised directly; the hardening paths compile-verified and reviewed (they need a staged install failure to exercise live).

## Notes

- Production signing stays the maintainer's Developer ID (unchanged). Fork testing uses a free self-signed cert (Squirrel.Mac checks a signature *match*, not notarization).
- "What's new" links to the GitHub release for now; an in-app release **news post** + **auto versioning/changelog** are deferred to **SOD-37**.

Refs SOD-18.

### DIFF
--- a/client/src/commonMain/kotlin/se/soderbjorn/lunamux/client/newsupdates/NewsUpdatesBackingViewModel.kt
+++ b/client/src/commonMain/kotlin/se/soderbjorn/lunamux/client/newsupdates/NewsUpdatesBackingViewModel.kt
@@ -123,10 +123,11 @@ fun createNewsUpdatesBackingViewModel(
  *   comparison baseline against the manifest's `latestVersionCode`.
  * @param currentVersionName the running build's human-readable version, logged
  *   for context (not used in the comparison).
- * @param enableUpdateCheck when false, the `versions.json` update comparison is
- *   skipped entirely (news still flows normally). Used by the desktop/Electron
- *   build, where electron-updater owns in-app updates and this manifest would
- *   otherwise double-notify. Defaults to true, so Android/iOS are unchanged.
+ * @param enableUpdateCheck when false, `versions.json` is not fetched and its
+ *   update comparison is skipped entirely (news still flows normally). Used by the
+ *   desktop/Electron build, where electron-updater owns in-app updates and this
+ *   manifest would otherwise double-notify. Defaults to true, so Android/iOS are
+ *   unchanged.
  * @param scope coroutine scope the check loop runs in; defaults to a background
  *   [SupervisorJob] scope so a failed check never tears down siblings.
  * @param manifestUrl the version manifest URL; overridable for testing.
@@ -288,7 +289,10 @@ class NewsUpdatesBackingViewModel(
             "NewsUpdates: checking $manifestUrl + $newsUrl for platform=$platformId " +
                 "code=$currentVersionCode name=$currentVersionName",
         )
-        val versions = fetchVersionManifest()
+        // Desktop disables the versions.json update path (electron-updater owns
+        // updates there), so skip the fetch entirely rather than spend a request on
+        // a manifest we'd ignore. News is always fetched.
+        val versions = if (enableUpdateCheck) fetchVersionManifest() else null
         val news = fetchNewsManifest()
         if (versions == null && news == null) {
             // Silent no-op on total failure: no state change, no timestamp advance.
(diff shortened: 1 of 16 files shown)


ottojaa:
https://github.com/user-attachments/assets/19d52b02-6d36-44e1-84dd-2882b958fc5e

^In action on my fork. Running the new release script creates a draft release in Github which contains all of the necessary files for the autoupdate to function -> publishing still happens by hand. Also good to know: the script can only be run by @soderbjorn because codesign will only work with your credentials for the time being

As a potential follow up, tangentially related [linear ticket](https://linear.app/soderbjorn/issue/SOD-37/automate-release-versioning-changelog-and-release-news-post) for potentially automating the whole changelog / news generation flow which is not very difficult (from my experience, libs like release-please or semantic-release do the hard lifting)

soderbjorn:
This looks like a super nice flow. Will check it out in next day or two.
By the way, I actually had Claude build me an issue tracker from scratch. Been babysitting it a lot and building it out bit by bit and it's almost ready to deploy. So I think I will actually move there shortly and stop with Linear. I have some other projects I also want to manage with it.

soderbjorn:
I have a separate branch where I've been interating on this flow, based on your work. I will let you know when I've come further!

PR #156, opened 2026-08-19
Make the Android overview an app switcher, with truthful previews and a dive transition

ottojaa:
Makes the Android overview behave like the OS app switcher: a card per tab, flung sideways over a dock that follows the row, pinchable between three card sizes, with a shared-element dive into the full-screen terminal — and previews that show the session's real screen rather than re-wrapped text.

Everything is in `androidApp/`, bar one field in the shared `LocalState` for the persisted zoom level. **No server changes**, and no changes to the vendored Termux modules.

### Review order

The commits are meant to be read in order, and the first three each stand on their own:

1. **Render terminal previews as the real screen** — `TerminalFrame`/`ThumbRun` + `snapshotFrame`, the registry publishing frames, `TerminalThumbnail` painting them, and the height-fill fit that gives preview/terminal parity.
2. **Dive into a terminal, and fly its content there** — `DiveTransition.kt`, the placeholder the flight actually flies, the content-box anchor, the crossfade handoff, and the placeholder pinned to the live view's geometry.
3. **Make the overview the app switcher** — `SwitcherCardRow` + `TabDock`, browse-≠-commit, the fling/settle tuning, the dock riding the drag, and the mid-settle tap fix.
4. **Fix what the review found** — a self-review pass over the three above: RIS/DECSCNM theme handling, an O(cols²) row walk rewritten to O(cols), a bounded registry (8 live entries, oldest unsubscribed retired first), and a dive-debounce that no longer depends on wall-clock time.
5. **Stop the tap watcher swallowing taps, and the dock caching an empty strip** — three regressions commit 4 introduced, caught on device: an overlay pointer modifier is not hit-transparent, so a watcher that declined a gesture swallowed it (dead taps on a resting card); the tap test asked how far the *row* had moved rather than the *finger*, so a flick dived; and the dock's scaled-width walk cached the answer for a not-yet-measured strip.
6. **Open the switcher zoomed out, and let a pinch change that** — the cards were 89% of the width, which is a size for reading one tab, not for seeing how many there are. The row now opens at the OS switcher's proportions and pinches between three levels (0.45 / 0.62 / 0.89 of the width), the old full-width card among them, persisted so a posture is chosen once. Details below.

### What to look at

- **Read-only invariant.** Previews never call `resize`/`send`, so a thumbnail can never shrink the shared PTY. `ScaleToBounds` is load-bearing for the same reason: the terminal end of the dive wraps an `AndroidView` whose relayout would vote a size to the server, so the flight only ever scales a layer.
- **A zoom level is a layout, not a permanent transform.** The cheap implementation — leave the row scaled by a `graphicsLayer` and keep one layout — is wrong here: shared elements take their bounds from the lookahead pass, which does not see an ancestor's layer, so a scaled row would fly the dive from a rect that is not the card you tapped. So the cards are *measured* at the level's size and only the pinch itself is a transform: it previews the next size while the cards' constraints hold still (nothing is remeasured or recomposed with fingers down), and the layout lands on it at lift. The commit changes the stride, so it re-centres via `requestScrollToItem` in the very measure pass that follows.
- **Cards scale in both dimensions, never width alone.** A thumbnail fills its box's height and crops the columns that overflow, so a card narrowed at full height would show *less* of the terminal the further you zoomed out. Uniform scaling leaves vertical slack instead, which carries the tab title — the thing that makes the smallest level navigable, and where the OS switcher puts the app name.
- **One layering wart.** `net/ConnectionHolder` imports `MiniTerminalRegistry` from `ui` to evict the frame cache on disconnect. Happy to invert it behind a listener if you'd rather.
- **Motion constants that look arbitrary and are not.** Settle stiffness 50 at damping 1.2, an 80 dp/s flick threshold, dock scale 0.78 / opacity 0.50 / gap 6dp, and the three zoom fractions: all chosen on the device against a screen recording of the OS switcher, via a slider panel that was removed once they settled. Each carries a comment saying so.
- **Main-thread emulator lock.** Leaving a terminal snapshots its screen synchronously under the emulator's monitor, because the flight starts in the same frame and an async hop cannot land in time. It is one pass over the screen grid; the price is waiting out at most one in-flight output chunk.

### Deliberately not included

- **`emitCell`/`isDefaultStyle` still duplicate the server's `GridSerializer`.** Sharing them needs a new file in the vendored `:terminal-core` plus server changes, and the O(cols) rewrite has since diverged the walk anyway.
- **Snapshot allocation is not optimised.** The frame DTO allocates ~1.6k short-lived objects per publish (~10⁵/s while flinging three cards). It is nursery garbage with no observed jank, and it replaces a registry that rebuilt the *entire* transcript string on every PTY event, unthrottled. `TerminalFrame.kt` names the two real mitigations and says to measure first. Zooming out attaches a card or two more, and so publishes a little more; still gated per attached card, and still worth measuring before touching.
- **The pinch is the only way to change zoom.** No control, no menu item — undiscoverable, but the level persists, so it is a once-ever gesture rather than a daily one. Easy to add an affordance if you disagree.
- **No unit tests**: `androidApp` has no test source set, and adding one is not this PR's job. Everything here was verified on device instead — several rounds of it.

### Device testing

Built and sideloaded repeatedly through development against a laptop server, on a passive-mirror session (the phone showing a laptop-width grid). Exercised: previews coming up themed on first attach and after reconnect; new pane / layout preset landing on the browsed card; fling, flick, edge-of-row and tap-during-settle; pinching between all three levels and the dive from each of them; the dive's reverse; a tab whose every pane is docked; and a terminal opened from list view, which has no cached frame and so still flies blank — expected, not a regression.

### DIFF
--- a/androidApp/src/main/kotlin/se/soderbjorn/lunamux/android/net/ConnectionHolder.kt
+++ b/androidApp/src/main/kotlin/se/soderbjorn/lunamux/android/net/ConnectionHolder.kt
@@ -7,6 +7,7 @@ import kotlinx.coroutines.withContext
 import kotlinx.coroutines.withTimeout
 import se.soderbjorn.lunamux.HostPort
 import se.soderbjorn.lunamux.android.BuildConfig
+import se.soderbjorn.lunamux.android.ui.MiniTerminalRegistry
 import se.soderbjorn.lunamux.client.CandidateConnection
 import se.soderbjorn.lunamux.client.CandidateConnector
 import se.soderbjorn.lunamux.client.ClientIdentity
@@ -259,6 +260,11 @@ object ConnectionHolder {
         currentWindowSocket = null
         currentClient?.close()
         currentClient = null
+        // The thumbnail frame cache outlives any registry (it is process-wide,
+        // so returning to the overview paints instantly), which means leaving a
+        // host would otherwise keep its rendered screens in memory and seed the
+        // next host's overview from them.
+        MiniTerminalRegistry.clearFrameCache()
     }
 
     /**
--- a/androidApp/src/main/kotlin/se/soderbjorn/lunamux/android/ui/DiveTransition.kt
+++ b/androidApp/src/main/kotlin/se/soderbjorn/lunamux/android/ui/DiveTransition.kt
@@ -1,0 +1,118 @@
+/**
+ * Shared-element "dive" transition plumbing between the overview's mini panes
+ * and the full-screen terminal.
+ *
+ * Tapping a terminal card in the overview should feel like diving into it: the
+ * card's bounds morph into the terminal screen's content box, and the reverse
+ * plays on back. This file carries the two composition locals the
+ * transition needs ([LunamuxApp] provides them: the app-level
+ * [SharedTransitionScope] around the NavHost and each involved destination's
+ * [AnimatedVisibilityScope]) and the [diveSharedBounds] modifier both ends
+ * attach.
+ *
+ * The experimental shared-transition opt-in is deliberately confined to this
+ * file and [LunamuxApp]; call sites only see a plain [Modifier] extension.
+ *
+ * IMPORTANT constraint: [diveSharedBounds] keeps `sharedBounds`' default
+ * `resizeMode = ScaleToBounds`, which measures the content once at its final
+ * bounds and only layer-scales during the flight. The terminal side wraps an
+ * `AndroidView` whose relayout fires size votes to the server
+ * (see `TerminalScreen`'s layout listener) — `RemeasureToBounds` would relayout
+ * it every frame of the animation and must never be used here.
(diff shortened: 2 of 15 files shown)


ottojaa:
@soderbjorn test it out a bit and let me know what you think!

I still haven't gotten around to setting up my Xcode sims, so no iOS yet.

soderbjorn (approved):

soderbjorn:
It's pretty nice! I like it! Feels like a step towards making the app a bit more solid. Thanks!

soderbjorn:
Nice with the transitions and stuff!

soderbjorn:
The fact that the windows are now proportional, makes it so that I might not see what is going on exactly without tapping them. But I think it's worth it, because the whole UX feels smoother and nicer.

PR #157, opened 2026-08-20
Make the Android terminal's native text actions work

ottojaa:
Long-pressing the terminal on Android raises the floating text-selection bar, and **every button on it is dead**. Two separate causes, which is why it looked like several bugs.

Tracked as LMX-158 (consolidating LMX-120 "paste noops" and LMX-121 "More noops").

## Why Copy and Paste did nothing

The Termux view never touches the clipboard itself — it calls its session's `onCopyTextToClipboard` / `onPasteTextFromClipboard`. `TerminalEmulatorHolder`'s session stubbed both out as `= Unit`.

The stub was there for a real reason, and the comment above it says so: the session passes `null` for Termux's client, so every callback must be overridden or it NPEs — `reset()` inside the emulator constructor already calls one. But two of them needed bodies rather than silence.

The same stub also swallowed **OSC 52**, so a program on the far end could not set the phone's clipboard either.

## Why "More…" did nothing

It called `showContextMenu()`, which needs a host overriding `onCreateContextMenu`. That is how Termux's activity-based UI is built; `MainActivity` is Compose and provides no such thing, so the item could never work on any device.

I removed it rather than giving the activity a context menu to satisfy one button. The bar is now **Copy and Paste** — both of which work, and no overflow. I did try Select all and Share in that slot; both were cut on review (a "Select all" that stops at the visible screen reads as a bug rather than a scope, and the handles already do that job; Share did not earn a button).

## The rest

- Paste goes through `TerminalEmulator.paste` rather than writing bytes directly, so bracketed-paste is honoured and the text lands on the ordinary write path — a paste into a pane the phone is only *mirroring* takes the PTY over first, exactly as a keystroke would.
- Below API 33 a short toast confirms a copy; from 33 up Android shows its own clipboard confirmation and ours would just duplicate it.
- New `TerminalClipboard` marshals to the main thread, because OSC 52 arrives on the emulator dispatcher while the selection bar arrives on the UI thread, and `ClipboardManager` is a UI-thread API in practice.

## Two more defects found in the same code

- **Paste stayed greyed out.** Built enabled-or-not once, with `onPrepareActionMode` returning `false` forever — so a selection begun with an empty clipboard kept a disabled Paste even after the user had copied something, a disabled button being indistinguishable from the dead one beside it. Now follows the clipboard via `OnPrimaryClipChangedListener` rather than polling: `render()` invalidates the action mode every frame and `hasPrimaryClip()` is a binder call, so a per-frame check would put a synchronous IPC in the draw path.
- **The bar anchored to the wrong place.** `onGetContentRect` multiplied out cell metrics by hand, making it the one pixel↔cell conversion that never learnt about the mirror's horizontal pan and vertical centring — the four in `TerminalView` carry a comment saying they must. Now goes through `getPointX`/`getPointY`.

## Removals worth a second look

Dropping "More…" orphaned `mStoredSelectedText`, whose whole purpose was that flow: stash the text before stopping the selection, so the handles do not float above the context menu that is about to need it. With no such flow, `getStoredSelectedText()` could only return null and `TerminalView.onContextMenuClosed()` could only clear nothing — and its javadoc pointed at the now-deleted `ACTION_MORE`. Both are gone, with three imports they were the last users of.

The one remaining `showContextMenu()` call — hardware-mouse right-click — is left behaving **exactly** as before, with a comment noting it is the same dead affordance. It is unreachable on a phone and not worth changing on a path I cannot test here.

## Deliberate: thumbnails get no clipboard

`appContext` is nullable, and `MiniTerminalRegistry` passes `null`. The overview's miniatures replay the same output stream as the real pane, OSC 52 included, so honouring it there would let merely *opening the overview* overwrite what the user copied — once per thumbnail of the session. They have no selection bar to paste from either.

## Testing

`:terminal-core:test` passes and the release APK builds. **Verified on device** (Pixel, sideloaded release build): copy, paste, and the paste-into-a-mirrored-pane take-over all behave.

Known and unrelated: Gboard's spacebar-slide cursor gesture does nothing in the terminal, because `onCreateInputConnection` reports `inputType = TYPE_NULL` and that gesture is built on the rich text-editing API. Reproduces in Termux for the same reason; filed separately as LMX-159 rather than folded in here.

### DIFF
--- a/androidApp/src/main/kotlin/se/soderbjorn/lunamux/android/ui/MiniTerminalRegistry.kt
+++ b/androidApp/src/main/kotlin/se/soderbjorn/lunamux/android/ui/MiniTerminalRegistry.kt
@@ -116,6 +116,12 @@ class MiniTerminalRegistry(
             // thumbnail sizes its emulator directly in the Size collector below.
             // A thumbnail never takes input, so the input handler is a no-op.
             serverGridPin = java.util.concurrent.atomic.AtomicReference(null),
+            // No clipboard for a thumbnail. It replays the same output stream as the
+            // real pane, OSC 52 included, so honouring a program's clipboard request
+            // here would let merely opening the overview overwrite what the user
+            // copied — once per thumbnail of the session. There is no selection bar
+            // to paste from either.
+            appContext = null,
             handleInput = {},
         )
         val emulator = createSyncedEmulator(session)
--- a/androidApp/src/main/kotlin/se/soderbjorn/lunamux/android/ui/TerminalClipboard.kt
+++ b/androidApp/src/main/kotlin/se/soderbjorn/lunamux/android/ui/TerminalClipboard.kt
@@ -1,0 +1,132 @@
+/**
+ * System-clipboard bridge for the Android terminal.
+ *
+ * The Termux view and emulator both hand clipboard work to their session's
+ * [com.termux.terminal.TerminalOutput] callbacks rather than touching
+ * [ClipboardManager] themselves: the text-selection action bar calls
+ * `onCopyTextToClipboard` / `onPasteTextFromClipboard`, and a program's OSC 52
+ * sequence calls `onCopyTextToClipboard` from inside the emulator. Lunamux's
+ * session (see [createExternalTerminalSession]) is the only implementation of
+ * those callbacks in this app, and this object is what it delegates to.
+ *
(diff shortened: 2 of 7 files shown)


soderbjorn (approved):

soderbjorn:
Great work! This part feels more stable now!

PR #158, opened 2026-08-20
Give the desktop icon transparent corners instead of white ones

ottojaa:
## The bug

The Dock, the app switcher and the DMG all draw a white frame around the rounded square of the app icon.

macOS does not mask an app icon — whatever the artwork paints in its corners is what you see — and this artwork painted them opaque white. `electron/icons/icon.png` was 1024×1024 RGBA with alpha 255 on every one of its 1 048 576 pixels, 44 068 of them pure white and all of those inside the four squircle corners. `icon.icns` and `icon.ico` are cut from the same master and carried the same frame.

Android was never affected: its launcher icon is the vector adaptive pair (`ic_launcher_background` + `ic_launcher_foreground`), not this PNG.

## The fix

Every corner pixel turned out to be an exact blend of white over the background `#222022` — not one pixel in the corner zones was off that line — so the alpha is recoverable from the blend ratio instead of having to re-cut the shape with a fresh mask. That keeps the original anti-aliased edge rather than approximating it, and it is verifiable: each repaired pixel composited back over white equals exactly what it was before, and nothing outside the corners changed at all.

`icon.icns` (full 16→1024 iconset via `iconutil`) and `icon.ico` (16/32/48/64/128/256) are regenerated from the repaired master, resized in premultiplied alpha so the transparent side cannot bleed a light fringe into the edge on the way down to 16px. Both keep the same entry sets and formats they had before, and `directories.buildResources: "icons"` is untouched, so electron-builder consumes them exactly as it did.

`server/src/main/resources/lunamux-icon.png` — the icon of the server's Swing settings and device-auth dialogs, loaded in `SettingsDialog.kt` and `DeviceAuth.kt` — is the same artwork and gets the same repair.

Left alone deliberately: the iOS app icon (the App Store rejects icons with an alpha channel, and iOS masks the artwork anyway, with a corner radius wider than this one) and the Play Store listing icon.

## Verification

Two throwaway `.app` bundles were built, one carrying the old `icon.icns` and one the new, and their icons read back with `NSWorkspace.shared.icon(forFile:)` — the same API the Finder, the Dock and the app switcher draw from:

| bundle | corner pixel | alpha range | near-white pixels |
| --- | --- | --- | --- |
| old `icon.icns` | `(255,255,255,255)` | 255–255 | 10 848 |
| new `icon.icns` | `(0,0,0,0)` | 0–255 | 0 |

Drawing the `.icns` files directly through `NSImage`, with no icon-services cache in the path, gives the same two answers: the old file reproduces the reported white frame, the new one has none. Confirmed visually in the Finder on the two bundles as well.

Also checked on the assets themselves: alpha spans 0–255 with a fully transparent corner pixel in both master PNGs, in all ten iconset entries inside the new `.icns` and in all six `.ico` frames, with zero near-white pixels anywhere; and diffed against the old master, 0 changed pixels outside the corner zones.

Not verified by a packaged build — that needs a full `electron-builder` run and a relaunch of the app, which stops the background server.

(Note that `qlmanage -t` cannot be used to check this: it flattens a document thumbnail onto white and reports white corners for both the old and the new file.)


ottojaa:
Alternative to Opus' PR description bible:

Before:

image


After:

image

soderbjorn (approved):

soderbjorn:
This has been something that annoyed me for a long time. For some reason, I only get it on the custom builds, the signed package shows fine for whatever reason. Great that you found the problem!

ottojaa:
> This has been something that annoyed me for a long time. For some reason, I only get it on the custom builds, the signed package shows fine for whatever reason. Great that you found the problem!

Hah yeah it annoyed me for weeks, but just not enough to do anything about it till now 😅

